PR #1, opened 2023-03-24
Basic multi-provider scaffold of test suites

AverageMarcus:
This PR adds the initial structure and (very) basic tests for each of the currently supported provides: CAPA, CAPZ and CAPVCD.

All providers have a "standard" test suite setup that currently just runs the common tests.

Provider teams are expected to add more common, cross-provider tests to the `common` set of tests so that they run against all provider WCs. They can then add more provider-specific tests into their own `provider/XXX` test suites.

The `standard` test suite is aimed to be the most default setup for a workload cluster that should "just work" for customers. Additional test suites can then be added as extra directories alongside the `standard` directory to account for variations in WCs, such as different network configuration, etc.

All tests are built using our [clustertest](https://github.com/giantswarm/clustertest) framework and Ginkgo.

The test suites must be run with the `E2E_KUBECONFIG` environment variable set to the path of a valid kubeconfig containing the expected context of each test suite (see README.md).

Tests can be run with something like:
```
E2E_KUBECONFIG=~/.kube/clusters/e2e.yaml ginkgo -v ./providers/capa/standard
```

**Notes:**
* Deletion of CAPVCD workload clusters is currently **broken** so the tests will currently fail during the `AfterSuite` stage. See https://github.com/giantswarm/giantswarm/issues/26422
* CAPVCD expects a secret containing credentials to exist in the namespace it uses for WCs so for now the CAPVCD tests don't use a random Org each time but instead use the `giantswarm` Org.
* There is currently no automated running of these tests. This will come a little while later after the multiple context PR mentioned above is released. You can follow this issue: https://github.com/giantswarm/roadmap/issues/2081

Fixes https://github.com/giantswarm/roadmap/issues/2221


providers/capz/standard/capz_suite_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	logger.LogWriter = GinkgoWriter
+
+	framework, err = clustertest.New()
+	if err != nil {
+		panic(err)
+	}
+
+	cluster = application.NewClusterApp(utils.GenerateRandomName("t"), application.ProviderAzure).
+		WithAppVersions("", "v0.0.13"). // TODO: Remove when default-apps-azure is made public
```
bavarianbidi, line 38: It's public right now

AverageMarcus:
Updated to use [v0.0.5](https://github.com/giantswarm/clustertest/releases/tag/v0.0.5) of clustertest that now supports (and enforces) the use of kubeconfig context. Each provider currently expects the provided kubeconfig to have one of the following context names specified: `capa`, `capz` and `capvcd`. Each of these should be associated with an appropriate MC for that provider. 

More specific context can be added in the future (e.g. a CAPA MC with private networking available) by defining a new context name to use for it.

providers/capa/standard/capa_suite_test.go
```diff
@@ -0,0 +1,63 @@
...
+		WithAppValuesFile(path.Clean("./test_data/cluster_values.yaml"), path.Clean("./test_data/default-apps_values.yaml"))
+
+	BeforeSuite(func() {
+		logger.Log("Workload cluster name: %s", cluster.Name)
+
+		applyCtx, cancelApplyCtx := context.WithTimeout(ctx, 20*time.Minute)
+		defer cancelApplyCtx()
+
+		client, err := framework.ApplyCluster(applyCtx, cluster)
+		Expect(err).ToNot(HaveOccurred())
```
yulianedyalkova, line 49: Just syntax sugar so feel free to ignore but I think `NotTo` is used a lot more than `ToNot`.

AverageMarcus: I actually got that suggestion from running `ginkgolinter` against the codebase. It seems it treats them both the same so I'm happy to switch them all :)

providers/capvcd/standard/capvcd_suite_test.go
```diff
@@ -0,0 +1,65 @@
...
+	ctx := context.Background()
+
+	logger.LogWriter = GinkgoWriter
+
+	framework, err = clustertest.New(KubeContext)
+	if err != nil {
+		panic(err)
+	}
+
+	cluster = application.NewClusterApp(utils.GenerateRandomName("t"), application.ProviderCloudDirector).
```
erkanerol, line 40: Which version is used here? How are we supposed to run this test suite? Per PR, Per release or periodically? And where? Sorry for my dumb questions.

AverageMarcus: > Which version is used here? 
When not specified, as with this, then the latest version is retrieved from GitHub. I've [just merged a PR](https://github.com/giantswarm/clustertest/pull/15) that will allow for overriding that also with an environment variable (e.g. so a CI pipeline can pass it in) but I haven't updated this PR yet.

>  How are we supposed to run this test suite?
Currently it's only manual / locally following the instruction in the PR description. Once I have _something_ available to work with I can hook it up to a Tekton pipeline and work on getting it triggered automatically on PRs. I have a [WiP PR](https://github.com/giantswarm/tekton-resources/pull/41) waiting on this one to implement that but it's currently untested and will require this fixing - https://github.com/giantswarm/giantswarm/issues/26422.

The general idea is that these will run, at least initially, for the following situations:
* New PRs opened to this repo to test changes to the tests themselves.
* New PRs to the `cluster-X` apps
* New PRs to the `default-apps-X` apps

Following that we'll likely also trigger it on some form of cron trigger to try and catch mistakes made to the apps deployed to the MCs.

Eventually we want to get to a point where we can trigger some tests from other app repos too but we're still a long way off from that.

AndiDog (approved):

PR #7, opened 2023-03-31
Prebuild test suites in Docker image

AverageMarcus:
The compilation of the tests is fairly slow in the docker image (especially considering how little code there currently is) so we should pre-compile them into the docker image to speed up the running stage in CI/CD.


calvix (approved):
:magic_wand:

PR #12, opened 2023-04-12
Run tests against an existing cluster

mnitchev:
### Description

Towards https://github.com/giantswarm/giantswarm/issues/26378

This has several notable changes:
- The `framework` and `cluster` vars are now global vars for the `common` package. This removes the need of having the BeforeSuite and AfterSuite be set up in the test entrypoint function, because now the global vars in `common` can be set after they are initiated (whenever that is). Currently they are passed in the `Run` function  which is executed before the BeforeSuite, since ginkgo will execute the Describe that calls it first.
- `LoadCluster` will now check for the E2E_WC_NAME and E2E_WC_NAMESPACE env vars and return a nil cluster if they aren't set. This isn't perfect, but it removes the need for all the provider suites to check these env vars themseves.
- Since we're loading an existing cluster, we want the user to be responsible for cleaning it up. The way it's written now, there is no way of knowing if a cluster is loaded or created by the tests in the AfterEach. To combat this the cleanup is moved to a `DeferCleanup` func after a cluster is created. This is less explicit perhaps, but at least there's no need to create some kind of flag for cleanup.
-   Instead of explicitly calling a function that sets up the common tests we can leverage the package that we anyway import. This removes the need to explicitly call the `common.Run` function in all the suites. Just importing the package will set up the tests. This has the downside of being less explicit. We have to make this clear somehow that these tests are set up if we go this route.

### References
- [Example ginkgo suite composition](https://github.com/onsi/composition-ginkgo-example)


mnitchev:
This change is needed before merging (also `go mod tidy` when it's available): https://github.com/giantswarm/clustertest/pull/20

fiunchinho (approved):
LGTM

common/common.go
```diff
@@ -9,23 +9,28 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-func Run(framework *clustertest.Framework, cluster *application.Cluster) {
+var (
+	Framework *clustertest.Framework
+	Cluster   *application.Cluster
+)
+
+var _ = Describe("Common tests", func() {
```
mnitchev, line 17: The downside to this change is that it's less explicit. You have to know that importing the package will set up the tests.
I've left this as a separate commit so we can revert it if we decide we don't want to go that way. Running just `common.Run()` with the now exported `Framework` and `Cluster` vars works

fiunchinho: I prefer the explicit version, but I'm not blocking this.

mnitchev: I think I'm also leaning towards the more explicit version. But I want to hear other people's opinions

AverageMarcus: I also think explicit as ordering might be needed.

providers/capa/standard/capa_suite_test.go
```diff
@@ -8,56 +8,73 @@ import (
...
+
+	common.Framework = framework
+	common.Cluster = cluster
+})
+
+func setUpWorkloadCluster() *application.Cluster {
+	cluster, err := framework.LoadCluster()
+	Expect(err).NotTo(HaveOccurred())
+	if cluster != nil {
+		return cluster
```
AverageMarcus, line 53: Might be worth logging out here that we've loaded an existing cluster and the name of that cluster.

providers/capa/standard/capa_test.go
```diff
@@ -1,11 +0,0 @@
-package standard
```
AverageMarcus, line 1: I think keeping these files around would be best as this is where the provider-specific tests would eventually go.

AverageMarcus (approved):
If it works I'm happy with it :) Just make sure to update readme with new details.

I do think the explicit running of the common tests should be kept though.

providers/capa/standard/capa_test.go
mnitchev: Done

providers/capa/standard/capa_suite_test.go
mnitchev: Added a log in there. Thanks

AverageMarcus:
@mnitchev 

```
failed to create task run pod "pr-cluster-test-suites-12-cluster-test-suitesghlrk-run-tests": translating TaskSpec to Pod: GET https://quay.io/v2/giantswarm/cluster-test-suites/manifests/1.0.2-b17fb01823fe2bf4c0174a4fc769c80f50986e0b: MANIFEST_UNKNOWN: manifest unknown; map[]. Maybe missing or invalid Task tekton-pipelines/run-wc-e2e-tests
```

Looks like you need to wait a bit so that the docker image is built and pushed.

PR #15, opened 2023-04-25
Add test checks all pods are running

mnitchev:
To do this we need to check that a condition is eventually consistent, i.e. an action **eventually** passes for a **period** of time. Ginkgo by default doesn't have a way to do this. We also need this because the pods could transition through states or new pods can show up and for just a moment all pods could end up Running or Successful, when just after a pod could fail or a new one could pop up. To solve this I added an exported function - `Consistent` - that will perform an action with specified number of attempts and polling period.

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -1,9 +1,15 @@
 package common
 
 import (
+	"context"
+	"fmt"
+	"time"
+
 	"github.com/giantswarm/clustertest"
 	"github.com/giantswarm/clustertest/pkg/application"
 	"github.com/giantswarm/clustertest/pkg/client"
+	"github.com/giantswarm/clustertest/pkg/wait"
+	corev1 "k8s.io/api/core/v1"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


calvix (approved):

mnitchev:
> @mnitchev
> 
> > Ginkgo by default doesn't have a way to do this
> 
> Are you aware of `Consistently`? See https://onsi.github.io/ginkgo/#patterns-for-asynchronous-testing
> 
> ```lisp
> Both Eventually and Consistently perform asynchronous assertions by polling the provided input. In the case of Eventually, Gomega polls the input repeatedly until the matcher is satisfied - once that happens the assertion exits successfully and execution continues. If the matcher is never satisfied Eventually will time out with a useful error message. Both the timeout and polling interval are [configurable](https://onsi.github.io/gomega/#eventually).
> 
> In the case of Consistently, Gomega polls the input repeatedly and asserts the matcher is satisfied every time. Consistently only exits early if a failure occurs - otherwise it continues polling until the specified interval elapses. This is often the only way to assert that something "does not happen" in an asynchronous system.
> ```

Yes, but it can't be used in the same way here. Only feasible way to write it is like this:
```go
		Eventually(func() error {
			Consistently(checkPodSuccessfulPhase(wcClient), "10s", "1s").Should(Succeed())
			return nil
		}).
			WithTimeout(wait.DefaultTimeout).
			WithPolling(wait.DefaultInterval).
			Should(Succeed())
```
But this won't work. Both `Eventually` and `Consistently` will fail the test if the matcher isn't satisfied. After the timeout expires for `Eventually` and on the first failure for `Consistently`. So in this case the the entire test will fail (not just the current iterration of the Eventually loop) as soon as there is a pod that isn't in a successful state.

fiunchinho (approved):

mnitchev:
> @mnitchev In my previous companies, we were wrapping `Eventually` and `Consistently` like
> 
> ```
> func EventuallyConsistent(assertFunction){
>   Eventually(assertFunction)
>   Consistently(assertFunction)
> }
> ```
> 
> Example https://github.com/kubevirt/kubevirt/blob/bbf4069c2ea6164b9853cd6d1ae55704c7d944b1/tests/hyperv_test.go#L179


@erkanerol From the example you posted: that means just running them one after the other?
That also won't work - when the cluster is starting up you have pods coming up all the time, but occasionally have a second or two where all the current pods are running, but then the next second there's another pod coming up. So what you have is the Eventually passing, but then Consistently failing because the new pods are coming up.

PR #19, opened 2023-05-04
Add common tests checking if nodes are in desired ranges

mnitchev:


mnitchev:
This will fail until a new version of `clustertest` is release with these changes https://github.com/giantswarm/clustertest/pull/27

common/common.go
```diff
@@ -46,6 +47,70 @@ func Run() {
...
+		Expect(err).NotTo(HaveOccurred())
+
+		Eventually(wait.Consistent(checkWorkerNodesReady(wcClient, values), 12, 5*time.Second)).
+			WithTimeout(wait.DefaultTimeout).
+			WithPolling(wait.DefaultInterval).
+			Should(Succeed())
+	})
+}
+
+func checkControlPlaneNodesReady(wcClient *client.Client, values *application.ClusterValues) func() error {
```
fiunchinho, line 74: Would it make sense to move these functions `checkControlPlaneNodesReady` and `checkWorkerNodesReady` to `clustertest`? I see them as something that the framework provides for the end users, that could be used in different suites.

nprokopic: +1 to what Jose suggests

mnitchev: The fremework already provides the generic `AreNumNodesReady` and `AreNumNodesReadyWithinRange`, so I wouldn't put these ones there because they're too specific and they're used currently in one place.
I think we can review this again later when we have more tests and we know what we want to expose.

fiunchinho (approved):

providers/capz/standard/capz_test.go
```diff
@@ -1,11 +1,83 @@
...
+
+		minNodes += pool.MinSize
+		maxNodes += pool.MaxSize
+	}
+	expectedNodes := wait.Range{
+		Min: minNodes,
+		Max: maxNodes,
+	}
+
+	workersFunc := wait.AreNumNodesReadyWithinRange(context.Background(), wcClient, expectedNodes, &cr.MatchingLabels{"node-role.kubernetes.io/worker": ""})
```
bavarianbidi, line 74: on `kubeadm` based clusters (like with CAPI), the only `node-role.kubernetes.io` label which gets set is `node-role.kubernetes.io/control-plane` for control-plane nodes.

normal worker nodes doesn't have this label. And even in the k/k code-base there are only two consts (`LabelNodeRoleControlPlane = "node-role.kubernetes.io/control-plane"` and some kind of `labelNodeRolePrefix = "node-role.kubernetes.io/"` which seems orphaned)

To have this very generic for all providers, wdyt about listing the worker nodes by searching where `node-role.kubernetes.io/control-plane` is not set?

```
k get nodes -l node-role.kubernetes.io/control-plane!=
NAME             STATUS   ROLES    AGE   VERSION
oa92854g1ae12c   Ready    <none>   9d    v1.22.9+vmware.1
oa92854g497959   Ready    <none>   9d    v1.22.9+vmware.1
oa92854gb47aae   Ready    <none>   9d    v1.22.9+vmware.1

k get nodes -l node-role.kubernetes.io/control-plane=
NAME             STATUS   ROLES                  AGE   VERSION
oa92854g1510c4   Ready    control-plane,master   9d    v1.22.9+vmware.1
oa92854g29a6c7   Ready    control-plane,master   9d    v1.22.9+vmware.1
oa92854g6587c5   Ready    control-plane,master   9d    v1.22.9+vmware.1
```

Also not sure if it's that useful to have this dedicated `worker` label set on all worker nodes. At least on a vanilla kubernetes you don't get this (e.g. a `kind` based cluster).

bavarianbidi: > All CAPI clusters should be running [giantswarm/capi-node-labeler-app](https://github.com/giantswarm/capi-node-labeler-app) which handles adding these labels.

didn't know that (and also the first time i've seen this app since i've joined GS :see_no_evil: )
BTW: this App isn't a thing  in CAPO,CAPVCD,CAPZ and/or CAPV  ( cc FYI @giantswarm/team-rocket / @giantswarm/team-clippy ).

And if it's mandatory to get the `cluster-test` running successfully it might get moved into the `cluster-*` app (like all the `CxI` + `CoreDNS` stuff @kopiczko has started with)

gawertm: if that is the case, then I think it would be good to discuss in Kaas Sync or area kaas channel?

AverageMarcus: Is there anything to discuss? We want our clusters labeled consistently don't we? And to follow on from vintage without any surprises. Is there a reason not to include it?

kopiczko: This is the first time I hear about this app too.

I'm not a fan of bloating `cluster-<provider>` apps without a good reason. You have to understand we make dependencies between teams deeper every time we have to add something there and it already looks pretty bad with the current teams structure.

It feels like we can get away with the out-of-the-box labelling or no? Is there a reason to not do that?

AverageMarcus: There is no out-of-the-box labelling of worker nodes. only control plane nodes.

If you want to discuss it then by all means lets talk in KaaS sync. I was just under the impression that it was already in place for all our CAPI clusters (based on the name) so that it minimizes migration pain from vintage clusters.

kopiczko:
```
$ k get node -l node-role.kubernetes.io/control-plane
NAME             STATUS   ROLES                  AGE   VERSION
oa92854g1510c4   Ready    control-plane,master   10d   v1.22.9+vmware.1
oa92854g29a6c7   Ready    control-plane,master   10d   v1.22.9+vmware.1
oa92854g6587c5   Ready    control-plane,master   10d   v1.22.9+vmware.1


$ k get node -l '!node-role.kubernetes.io/control-plane'
NAME             STATUS   ROLES    AGE   VERSION
oa92854g1ae12c   Ready    <none>   10d   v1.22.9+vmware.1
oa92854g497959   Ready    <none>   10d   v1.22.9+vmware.1
oa92854gb47aae   Ready    <none>   10d   v1.22.9+vmware.1
```

If you can select workers (without the label) using kubectl then I guess it's also possible directly in Go code.

AverageMarcus: I know, but that is then a change our customers need to make to their workloads.

kopiczko: I'm not saying we shouldn't install capi-node-labeler-app with default apps, but for the test we could rely only on node-role.kubernetes.io/control-plane label

AverageMarcus: Oh, yeah I see what you mean. That would work too 😄 If we do install the labeler on all providers then it also provides an extra test that that app worked as expected.

fiunchinho (approved):

AverageMarcus (approved):

PR #26, opened 2023-06-05
Add DNS tests

mnitchev:
### What this PR does

Tests the common DNS records created for WCs

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/common/dns.go
+++ b/common/dns.go
@@ -17,7 +17,7 @@ func runDNS() {
 	Context("dns", func() {
 		var (
 			resolver *net.Resolver
-			values   *application.DefaultAppsValues
+			values   *application.ClusterValues
 		)
 		getARecords := func(domain string) ([]net.IP, error) {
 			records, err := resolver.LookupIP(context.Background(), "ip", domain)
@@ -31,7 +31,7 @@ func runDNS() {
 		}
 
 		BeforeEach(func() {
-			values = &application.DefaultAppsValues{}
+			values = &application.ClusterValues{}
 			err := Framework.MC().GetHelmValues(Cluster.Name, Cluster.Namespace, values)
 			Expect(err).NotTo(HaveOccurred())
 
--- a/common/dns.go
+++ b/common/dns.go
@@ -61,7 +61,7 @@ func runDNS() {
 		})
 
 		It("sets up the bastion DNS records", func() {
-			bastionDomain := fmt.Sprintf("bastion1.%s", values.BaseDomain)
+			bastionDomain := fmt.Sprintf("bastion1.%s.%s", Cluster.Name, values.BaseDomain)
 			var records []net.IP
 			Eventually(func() error {
 				var err error
--- a/common/dns.go
+++ b/common/dns.go
@@ -21,8 +21,7 @@ func runDNS() {
 
 		BeforeEach(func() {
 			values = &application.DefaultAppsValues{}
-			defaultAppsName := fmt.Sprintf("%s-default-apps", Cluster.Name)
-			err := Framework.MC().GetHelmValues(defaultAppsName, Cluster.Namespace, values)
+			err := Framework.MC().GetHelmValues(Cluster.Name, Cluster.Namespace, values)
 			Expect(err).NotTo(HaveOccurred())
 
 			resolver = &net.Resolver{
--- a/common/dns.go
+++ b/common/dns.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"github.com/giantswarm/clustertest/pkg/application"
-	"github.com/miekg/dns"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
@@ -21,8 +20,7 @@ func runDNS() {
 
 		BeforeEach(func() {
 			values = &application.DefaultAppsValues{}
-			defaultAppsName := fmt.Sprintf("%s-default-apps", Cluster.Name)
-			err := Framework.MC().GetHelmValues(defaultAppsName, Cluster.Namespace, values)
+			err := Framework.MC().GetHelmValues(Cluster.Name, Cluster.Namespace, values)
 			Expect(err).NotTo(HaveOccurred())
 
 			resolver = &net.Resolver{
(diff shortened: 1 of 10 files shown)


AverageMarcus (approved):
LGTM assuming tests pass

common/dns.go
```diff
@@ -0,0 +1,82 @@
...
+			dnsClient := new(dns.Client)
+			var dnsResponse *dns.Msg
+			Eventually(func() error {
+				var err error
+				dnsResponse, _, err = dnsClient.Exchange(dnsMessage, "8.8.8.8:53")
+				return err
+			}).Should(Succeed())
+
+			Expect(dnsResponse.Answer[0].(*dns.CNAME).Target).To(Equal(ingressDomain))
+		})
```
fiunchinho, line 80: I think this changed recently for CAPZ. I remember something about CAPZ not using the wildcard anymore and using specific ingresses instead. You would have to test if this works in CAPZ.

common/dns.go
```diff
@@ -0,0 +1,82 @@
...
+func runDNS() {
+	Context("dns", func() {
+		var (
+			resolver *net.Resolver
+			values   *application.DefaultAppsValues
+		)
+
+		BeforeEach(func() {
+			values = &application.DefaultAppsValues{}
+			defaultAppsName := fmt.Sprintf("%s-default-apps", Cluster.Name)
```
fiunchinho, line 24: Any reason to take the value from the `default-apps-$provider` app instead of the `cluster-$provider` app? I'm afraid that if we keep moving apps from the `default-apps` app into the `cluster` app, maybe we reach a point where we don't need the `baseDomain` value in there anymore, and it'd only be present in the cluster app.

fiunchinho (changes requested):
A few questions

common/dns.go
mnitchev: No specific reason. I'll change it to the cluster app

common/dns.go
bavarianbidi: Yes, `CAPZ` doesn't have the wildcard (xref: https://github.com/giantswarm/roadmap/issues/1037#issuecomment-1431037446)

common/dns.go
```diff
@@ -0,0 +1,89 @@
...
+				Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+					d := net.Dialer{
+						Timeout: time.Millisecond * time.Duration(10000),
+					}
+					return d.DialContext(ctx, "udp", "8.8.4.4:53")
+				},
+			}
+		})
+
+		It("sets up the api DNS records", func() {
```
bavarianbidi, line 41: just FYI: as long as no "private" Clusters got created in `CAPZ`, everything is fine.
But if the WC is a private one, depending where these tests are run, if it's in the MC itself, the Record might be a different one - https://github.com/giantswarm/dns-operator-azure#architecture

njuettner:
@mnitchev any ideas when you can pick it up again 🙏🏻?

mnitchev:
I've removed the wildcard and ingress domains. These can be checked later with a test app that uses the ingress. Please re-review

common/dns.go
```diff
@@ -0,0 +1,65 @@
...
+			apiDomain := fmt.Sprintf("api.%s", values.BaseDomain)
+			var records []net.IP
+			Eventually(func() error {
+				var err error
+				records, err = resolver.LookupIP(context.Background(), "ip", apiDomain)
+				return err
```
AverageMarcus, line 45: I think we could do with some logging in here to show that something is still happening. The logs make it look like this test just got stuck.

common/dns.go
```diff
@@ -0,0 +1,65 @@
...
+			Expect(err).NotTo(HaveOccurred())
+
+			resolver = &net.Resolver{
+				PreferGo:     true,
+				StrictErrors: true,
+				Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+					d := net.Dialer{
+						Timeout: time.Millisecond * time.Duration(10000),
+					}
+					return d.DialContext(ctx, "udp", "8.8.4.4:53")
```
AverageMarcus, line 34: This wont work for private clusters, right? The API server endpoint wont resolve correctly.

vxav: I think you're right, if internet traffic is only allowed through HTTP proxy then port 53 will get dropped.

AverageMarcus: I was actually thinking more that Google DNS will return the wrong IP address for `api.xxxxx` compared to our VPN DNS server. But yeah, that also might be the case, I'm not sure if we allow DNS by default or not.

mnitchev: We have an issue to add private cluster tests here https://github.com/giantswarm/roadmap/issues/2614
I suggest we look at solutions then.

vxav: In theory the public record should be correct as it should resolve the actual IP of the cluster anyway that would only be reachable through the VPN. But then I don't know if other providers do it differently

AverageMarcus: No, at least not on AWS. We use private Route53 which isn't available publicly. (If I'm remembering correctly). The external will return the wildcard IP which isn't the right one.

vxav: Interesting I didn't know that, we create records in the "normal" route53.

mnitchev:
/run cluster-test-suites

Hopefully for the last time

common/dns.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/giantswarm/clustertest/pkg/wait"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func runDNS() {
+	Context("dns", func() {
+		var (
+			resolver *net.Resolver
+			values   *application.DefaultAppsValues
```
fiunchinho, line 20: Shouldn't this be `ClusterValues` ?

AverageMarcus:
The CAPA dns took nearly 30 minutes before it was available. The CAPZ one was immediately available. 🤔

common/dns.go
mnitchev: Yeah missed it, still worked though :laughing:

mnitchev:
/run cluster-test-suites

The new release of cluster-aws should solve the previous problem on CAPA

common/dns.go
jkremser: I think you are mixing "private hosted zone" and "public hosted zone" with the route53 api as such. The Route53 api is always publicly available to my knowledge but hosted zones can be private.

jkremser (approved):

PR #27, opened 2023-06-14
Update cluster values for cluster-cloud-director

erkanerol:
### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/providers/capvcd/standard/test_data/cluster_values.yaml
+++ b/providers/capvcd/standard/test_data/cluster_values.yaml
@@ -10,14 +10,23 @@ cloudDirector:
   ovdc: vDC 73640
   site: https://vmware.ikoula.com
   ovdcNetwork: capvcd-192.168.52.0
+
+cloudProvider:
+  enableVirtualServiceSharedIP: false
+  oneArm:
+    enabled: true
+        
 controlPlane:
   catalog: giantswarm
   replicas: 3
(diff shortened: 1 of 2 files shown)


providers/capvcd/standard/test_data/cluster_values.yaml
```diff
@@ -10,14 +10,23 @@ cloudDirector:
...
   sizingPolicy: m1.large
   template: ubuntu-2004-kube-v1.22.9
-network:
-  loadBalancer:
-    vipSubnet: 178.170.32.1/24
+
+connectivity:
+  network:
+    loadBalancers:
+      vipSubnet: "178.170.32.1/24"
```
AverageMarcus, line 28: Can this same value be used by two clusters at the same time or is it the same issue as CAPV?

vxav: This is a CAPVCD thing, if it is in the CAPV values it's a leftover from the capvcd=>capv port.

In CAPVCD specifically, this value stays the same across clusters (for the same infra environment).

AverageMarcus: Great.  just wanted to make sure this didn't need to be unique per cluster!

AverageMarcus (approved):

PR #28, opened 2023-06-16
PVC test

njuettner:
### What this PR does

Issue: https://github.com/giantswarm/giantswarm/issues/22739

### Checklist

### Trigger e2e tests

/run cluster-test-suites


common/common.go
```diff
@@ -112,3 +134,123 @@ func checkAllPodsSuccessfulPhase(wcClient *client.Client) func() error {
 		return nil
 	}
 }
+
+func createPodWithPVC(wcClient *client.Client) func() error {
```
AverageMarcus, line 138: You might want to take a look at how Mario has started structuring the code in https://github.com/giantswarm/cluster-test-suites/pull/26. It might be best to pull these PVC tests out in a similar way.

common/common.go
```diff
@@ -112,3 +134,123 @@ func checkAllPodsSuccessfulPhase(wcClient *client.Client) func() error {
...
+	return func() error {
+		// ensure we have at least one storage class available
+		storageClasses := &storagev1.StorageClassList{}
+		err := wcClient.List(context.Background(), storageClasses)
+		if err != nil {
+			return err
+		}
+		if len(storageClasses.Items) == 0 {
+			return fmt.Errorf("no storage classes found")
+		}
```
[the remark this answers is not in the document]
njuettner: good point, yeah that make sense.

common/common.go
```diff
@@ -112,3 +134,123 @@ func checkAllPodsSuccessfulPhase(wcClient *client.Client) func() error {
...
+
+		pvc := &corev1.PersistentVolumeClaim{
```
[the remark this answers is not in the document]
njuettner: yes currently the default SC we use for Azure and AWS, we use reclaim policy delete.
but to be sure I had another check which ensures the deletion of the pod/pvc.

common/common.go
njuettner: Thanks, I'll have a look

njuettner: I think it would make sense to merge https://github.com/giantswarm/cluster-test-suites/pull/26 first. Afterwards I'll rebase my PR 🙂

AverageMarcus: I agree! (cc @mnitchev)

common/storage.go
```diff
@@ -0,0 +1,178 @@
...
+
+		It("has created a pod with a pvc and the pvc is bound", func() {
+			Eventually(wait.Consistent(createPodWithPVC(wcClient), 10, time.Second)).
+				WithTimeout(wait.DefaultTimeout).
+				WithPolling(wait.DefaultInterval).
+				Should(Succeed())
+		})
+
+		It("has deleted a pod with a pvc", func() {
+			Eventually(wait.Consistent(deletePodWithPVC(wcClient), 10, time.Second)).
```
mnitchev, line 54: This couples this test to the one in line 46 together. Tests should be independent from one another.  In this case if the tests are run in parallel or just out of order, the delete will fail and the pod will be leaked.

common/storage.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+		if len(storageClasses.Items) == 0 {
+			return fmt.Errorf("no storage classes found")
+		}
+
+		return nil
+	}
+}
+
+func createPodWithPVC(wcClient *client.Client) func() error {
```
mnitchev, line 78: Nitpick. It would probably be nicer to have these manifests in yaml in an `assets` directory. In general I think we should have an test app that we deploy on the cluster once that uses all the features like PVC that we want to test, but we can move in that direction later.

common/storage.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func deletePodWithPVC(wcClient *client.Client) func() error {
+	return func() error {
+
+		pod := &corev1.Pod{}
+		err := wcClient.Get(context.Background(), types.NamespacedName{Name: pvcPodName, Namespace: corev1.NamespaceDefault}, pod)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return nil
```
mnitchev, line 151: We should always try and delete both here instead. Also there's no point in getting the pod or PVC before deleting. IIRC `Delete` only needs the type and name/namespace. We can ignore the `IsNotFound` error for the delete and then move on to the PVC.

mnitchev (changes requested):
Requested some changes.

njuettner:
@mnitchev I addressed you comments, unfortunately the tests are not running properly due to the machine pool bug in CAPI 1.4.0

do you still could take again? I think the code should be better now 🙏🏻.

common/common.go
```diff
@@ -17,4 +17,6 @@ var (
 func Run(cfg *TestConfig) {
 	runBasic()
 	runDNS(cfg.BastionSupported)
+	runStorage()
+	cleanupStorage()
```
AverageMarcus, line 21: This won't guarantee the order these are run.

common/storage.go
```diff
@@ -0,0 +1,177 @@
...
+		})
+
+		It("has deleted all objects in namespace test-storage", func() {
+			Eventually(wait.Consistent(deleteStorage(wcClient, namespace), 10, time.Second)).
+				WithTimeout(wait.DefaultTimeout).
+				WithPolling(wait.DefaultInterval).
+				Should(Succeed())
+		})
+	})
+}
```
AverageMarcus, line 91: Is this doing anything that's not already covered on line 54?

njuettner: Yeah you're right, removed it. Thanks <3

common/common.go
njuettner: Removed the cleanupStorage 👍🏻

mnitchev (approved):
LGTM as long as the tests pass

PR #34, opened 2023-07-13
Enable capv tests & let the IPAM in gcapeverde assign the IP for wc's control plane

jkremser:
### Checklist

### Trigger e2e tests

/run cluster-test-suites


providers/capv/standard/test_data/cluster_values.yaml
```diff
@@ -26,12 +26,13 @@ connectivity:
...
+    allowAllEgress: true
     controlPlaneEndpoint:
-      host: "10.10.222.231"  # [string] Manually select an IP for kube API.
+      host: ""  # [string] Manually select an IP for kube API. "" for auto selection from the ipPoolName pool.
       port: 6443
+      ipPoolName: "wc-cp-ips"  # [string] Ip for control plane will be drawn from this GlobalInClusterIPPool.
     loadBalancers:
-      cidrBlocks:
-      - "10.10.222.232/30"
+      cidrBlocks: []
```
AverageMarcus, line 35: Can we remove this as it's the default in the `values.yaml` ?

providers/capv/standard/test_data/cluster_values.yaml
```diff
@@ -26,12 +26,13 @@ connectivity:
   bastion:
     enabled: true
   network:
+    allowAllEgress: true
```
AverageMarcus, line 29: Just curious, why isn't this the default?

AverageMarcus (approved):
Looks good to me! :D Just a couple small comments but not blocking.

providers/capv/standard/test_data/cluster_values.yaml
jkremser: I'd go with something that works and perhaps remove / set this to false once all the apps have the required policies (NetworkPolicy & CiliumNetworkPolicy) properly bundled. I am still hitting cases where it's not there yet. Today I hit that again with service for validating&mutating webhooks for CRD.

AverageMarcus: Ah gotcha! It's for the Cilium cluster policy? I mistakenly thought it was some vsphere thing. Makes sense.

providers/capv/standard/test_data/cluster_values.yaml
jkremser: Should be gone. However this is not solved yet for testing, this IP range basically tells kubevip cloud provider that's running in WC, what IPs it can allocate for Services with type LoadBalancer. We need at least one IP for the ingress controller and the rest can be done using Ingress (grafana, dex and what not). So once/if the tests start to do something smart like checking if service A is reachable by dns name, it will probably fail.

It's my next thing to work on, though.

jkremser: our schema is not happy about it:

```
k describe app t-oyeoqr6 -n org-giantswarm | grep -A4 Reason                                                                                                                                ○ gcapeverde-admin@gcapeverde
    Reason:         values don't meet the specifications of the schema(s) in the following chart(s):
cluster-vsphere:
- connectivity.network.loadBalancers.cidrBlocks: Array must have at least 1 items

    Status:  not-installed
```
so putting this back

AverageMarcus: Can't you remove both the `controlPlaneEndpoint` and `loadBalancers` blocks as they already have the same defaults baked into the chart.

AverageMarcus: Oh, sorry I follow now. I thought the default values had a CIDR specified but I was looking at the wrong bit.

jkremser:
the tests are failing on this:
```
Common tests dns [It] sets up the bastion DNS records
/app/common/dns.go:64

  [FAILED] Timed out after 3600.001s.
  Expected success, but got an error:
      <*net.DNSError | 0xc0001c1d00>: 
      lookup [bastion1.t-bbrs7km.test.gigantic.io](http://bastion1.t-bbrs7km.test.gigantic.io/) on 172.31.0.10:53: no such host
      {
          Err: "no such host",
          Name: "[bastion1.t-bbrs7km.test.gigantic.io](http://bastion1.t-bbrs7km.test.gigantic.io/)",
          Server: "172.31.0.10:53",
          IsTimeout: false,
          IsTemporary: false,
          IsNotFound: true,
      }
  In [It] at: /app/common/dns.go:73 @ 07/14/23 12:24:52.897
```

Bastion hasn't been implemented for capv yet, is there a way to disable this test for vsphere environment? I've noticed in `cluster_values.yaml` there is:

```
connectivity:
  bastion:
    enabled: true
```
This is not read anywhere atm, would setting it to `false` made the test suite to not test it though? cc @mnitchev

PR #54, opened 2023-08-08
Add app tests

fiunchinho:
### Checklist

### Trigger e2e tests

/run cluster-test-suites


jkremser (approved):

fiunchinho:
> the deployed app does not mean the app si actually running or working, do we wanna cover a test scenario where we check the status of the app itself somehow?

I think testing all the behaviors from all the default apps from these e2e tests would be a huge effort. I believe it would be better to try to make charts have the `failed` status if the app is not working (i.e some `Job` failed, or `Deployment` is not `Ready`, etc).

Having said that, I also think it would be cool to have one test that deploys the ingress controller app together with a "helloworld" app that uses the `external-dns` annotation to create a DNS record and creates a `Certificate` using `cert-manager`. The test would then try to hit the created DNS using https. With that test we'd be testing many different components on a single test: aws authentication, cert-manager, ingress, external-dns...
But I'd do that on a separate PR.

common/apps.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/onsi/gomega/format"
+	"github.com/onsi/gomega/types"
+	ctrl "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func runApps() {
+	Context("default apps", func() {
+		var customFormatterKey format.CustomFormatterKey
+		BeforeEach(func() {
+			customFormatterKey = format.RegisterCustomFormatter(func(value interface{}) (string, bool) {
```
fiunchinho, line 19: Needed to have a better output when test fails. This is what it looks like

image

PR #57, opened 2023-08-09
Add hello-world app test

fiunchinho:
### What this PR does
Towards https://github.com/giantswarm/giantswarm/issues/22741

Here we are adding a basic test that deploys our hello-world app and tries to reach it via `https`. This test is useful because to be able to do that, the hello-world app needs a `cert-manager` `Certificate`, and the DNS record for the ingress controller needs to exist, so we make sure that `external-dns` is also working.

### Checklist

### Trigger e2e tests

/run cluster-test-suites


README.md
```diff
@@ -70,6 +70,7 @@ users:
   user:
     client-certificate-data: [REDACTED]
     client-key-data: [REDACTED]
+
```
AverageMarcus, line 73:
```suggestion
```

common/common.go
```diff
@@ -5,8 +5,9 @@ type TestConfig struct {
 }
 
 func Run(cfg *TestConfig) {
-	runBasic()
 	runApps()
+	runBasic()
```
AverageMarcus, line 9: I hope the order here’s isn’t important. It’ll not be guaranteed once parallel tests are implemented.

fiunchinho: I just ordered it alphabetically.

common/hello.go
```diff
@@ -0,0 +1,103 @@
...
+				return checkAppStatus(app)
+			}).
+				WithTimeout(3 * time.Minute).
+				WithPolling(5 * time.Second).
+				Should(Succeed())
+		})
+
+		It("hello world app responds successfully", func() {
+			Eventually(func() (*http.Response, error) {
+				return http.Get(fmt.Sprintf("https://hello-world.%s.gaws.gigantic.io", state.GetCluster().Name))
```
calvix, line 56: could we read the domain from somewhere like from cluster-values CM?

common/hello.go
```diff
@@ -0,0 +1,103 @@
...
+func runHelloWorld() {
+	Context("hello world", func() {
+		var nginxApp, helloApp *v1alpha1.App
+		var nginxConfigMap, helloConfigMap *v1.ConfigMap
+
+		BeforeEach(func() {
+			ctx := context.Background()
+			org := organization.New("giantswarm")
+
+			nginxApp, nginxConfigMap = deployApp(ctx, "ingress-nginx", "kube-system", org, "3.0.0", "./test_data/nginx_values.yaml")
```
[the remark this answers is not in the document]
AndiDog: Maybe [getLatestVersion from kubectl-gs](https://github.com/search?q=repo%3Agiantswarm%2Fkubectl-gs+%22func+getLatestVersion%22&type=code) could be helpful to get the latest version from the app catalog.

common/hello.go
AndiDog: Same in `providers/capa/standard/test_data/helloworld_values.yaml`

common/hello.go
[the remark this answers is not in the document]
AverageMarcus: If you want to use latest version then just don't supply a version when building the Application below. The framework will call out to GitHub to get the latest release.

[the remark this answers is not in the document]
AverageMarcus: See docs: https://pkg.go.dev/github.com/giantswarm/clustertest@v0.3.0/pkg/application#hdr-App_Versions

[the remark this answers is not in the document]
fiunchinho: I'm glad you asked me about this :smile: This is one of the many workarounds we need to do to make it work. If I don't pass any version value, the latest will be used as @AverageMarcus said. But the framework currently is not prepared (I think?) to handle apps that use a different name than the repository where they are hosted, meaning that the script fails to find the latest tag. By specifying a version, we work around that problem until passing a different name for the app and the repo is supported. I'll write a comment in the code explaining it.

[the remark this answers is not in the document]
AverageMarcus: Do you think we might be able to handle that usecase? The `Application` could maybe have a `Repo` property that defaults to the same as the name but could be user-set?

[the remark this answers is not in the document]
fiunchinho: yeah, I think that would work.

common/hello.go
fiunchinho: Done

providers/capz/standard/capz_test.go
```diff
@@ -23,7 +23,8 @@ type ClusterValues struct {
 
 var _ = Describe("Common tests", func() {
 	common.Run(&common.TestConfig{
-		BastionSupported: true,
+		BastionSupported:     true,
+		ExternalDnsSupported: false,
```
fiunchinho, line 27: We can enable it when this is merged https://github.com/giantswarm/default-apps-azure/pull/150

common/hello.go
```diff
@@ -0,0 +1,190 @@
...
+
+func runHelloWorld(externalDnsSupported bool) {
+	Context("hello world", func() {
+		var nginxApp, helloApp *v1alpha1.App
+		var nginxConfigMap, helloConfigMap *v1.ConfigMap
+		var helloWorldIngressHost string
+
+		BeforeEach(func() {
+			if !externalDnsSupported {
+				Skip("Bastion is not supported.")
```
AndiDog, line 31:
```suggestion
				Skip("external-dns is not supported")
```

common/hello.go
```diff
@@ -0,0 +1,190 @@
...
+			}
+
+			ctx := context.Background()
+			org := state.GetCluster().Organization
+
+			// The hello-world app ingress requires a `Certificate` and a DNS record, so we need to make sure `cert-manager` and `external-dns` are deployed.
+			Eventually(func() error {
```
AndiDog, line 38: [optional]

This block is quite repetitive. At some point, we may want to make this a helper taking only the app name and namespace.

common/hello.go
```diff
@@ -0,0 +1,190 @@
...
+				if err != nil {
+					return err
+				}
+				return checkAppStatus(app)
+			}).
+				WithTimeout(3 * time.Minute).
+				WithPolling(5 * time.Second).
+				Should(Succeed())
+
+			// `external-dns` will only create dns records for Ingresses on the `kube-system` namespace, that's why we install the nginx app in that namespace.
```
AndiDog, line 60: [minor]

Can we be specific here, since this may change in the future? I think this refers to the default value `namespaceFilter: kube-system` in external-dns-app?

common/hello.go
```diff
@@ -0,0 +1,190 @@
...
+			}).
+				WithTimeout(3 * time.Minute).
+				WithPolling(5 * time.Second).
+				Should(Succeed())
+
+			helloWorldIngressHost = fmt.Sprintf("hello-world.%s", getWorkloadClusterDnsZone(ctx))
+			helloAppValues := map[string]string{"IngressUrl": helloWorldIngressHost}
+			helloApp, helloConfigMap = deployApp(ctx, "hello-world", "giantswarm", org, "2.0.0", "./test_data/helloworld_values.yaml", helloAppValues)
+			Eventually(func() error {
+				// The hello-world app creates Ingress resources, and they won't be able to be created until the nginx webhooks are up and running, so the first time we try to install the hello-world app, it will fail.
```
AndiDog, line 77: > they won't be able to be created until

Please specify _why_, or else we'll never be able to remove this workaround anymore

common/hello.go
```diff
@@ -0,0 +1,190 @@
...
+				return checkAppStatus(app)
+			}).
+				WithTimeout(6 * time.Minute).
+				WithPolling(5 * time.Second).
+				Should(Succeed())
+		})
+
+		It("hello world app responds successfully", func() {
+			if !externalDnsSupported {
+				Skip("Bastion is not supported.")
```
AndiDog, line 103:
```suggestion
				Skip("external-dns is not supported")
```

common/hello.go
```diff
@@ -0,0 +1,190 @@
...
+			}
+
+			Eventually(func() (*http.Response, error) {
+				helloWorldIngressUrl := fmt.Sprintf("https://%s", helloWorldIngressHost)
+				logger.Log("Trying to get a successful response from %s", helloWorldIngressUrl)
+				return http.Get(helloWorldIngressUrl)
+			}, "10m", "5s").Should(HaveHTTPStatus(http.StatusOK))
```
AndiDog, line 110: There's the nginx-ingress _default backend_ which might also respond with HTTP 200, but that would mean a problem since the request didn't reach the Hello World app. Let's please check the content of the response to contain a certain string. [This](https://github.com/giantswarm/helloworld/blob/master/content/index.html) seems to be what the app would return.

common/hello.go
```diff
@@ -0,0 +1,190 @@
...
+		})
+
+		AfterEach(func() {
+			if !externalDnsSupported {
+				Skip("Bastion is not supported.")
```
AndiDog, line 115:
```suggestion
				Skip("external-dns is not supported")
```

common/hello.go
```diff
@@ -0,0 +1,190 @@
...
+func deployApp(ctx context.Context, name, namespace string, organization *organization.Org, version, valuesFile string, values map[string]string) (*v1alpha1.App, *v1.ConfigMap) {
+	var err error
+	managementClusterKubeClient := state.GetFramework().MC()
+	appBuilder := application.New(fmt.Sprintf("%s-%s", state.GetCluster().Name, name), name).
+		WithCatalog("giantswarm").
+		WithOrganization(*organization).
+		// If we don't pass any version value when building the app, the latest version will be calculated and used. But the calculation will fail when the app name doesn't match the repository name.
+		WithVersion(version).
+		WithInCluster(false).
+		// We need to manually set this label that should be automatically added once we can set the `app.Config` property.
```
AndiDog, line 144: Do we have an issue for `app.Config`?

common/hello.go
```diff
@@ -0,0 +1,190 @@
...
+
+	clusterValuesConfigMapData := make(map[string]interface{})
+	err = yaml.Unmarshal([]byte(clusterValuesConfigMap.Data["values"]), &clusterValuesConfigMapData)
+	Expect(err).ShouldNot(HaveOccurred())
+
+	if clusterValuesConfigMapData["baseDomain"] == "" {
+		Fail("baseDomain field missing from cluster-values configmap")
+	}
+
+	return fmt.Sprintf("%s", clusterValuesConfigMapData["baseDomain"])
```
AndiDog, line 189:
```suggestion
	return clusterValuesConfigMapData["baseDomain"]
```

providers/capa/standard/test_data/helloworld_values.yaml
```diff
@@ -0,0 +1,16 @@
+clusterName: {{ .ClusterName }}
+organization: {{ .Organization }}
+ingress:
+  annotations:
+    # kubernetes.io/ingress.class: nginx
```
AndiDog, line 5: [repeated]

Remove?

providers/capv/standard/capv_test.go
```diff
@@ -8,6 +8,7 @@ import (
 
 var _ = Describe("Common tests", func() {
 	common.Run(&common.TestConfig{
-		BastionSupported: false,
+		BastionSupported:     false,
+		ExternalDnsSupported: false,
```
AndiDog, line 12: Maybe something like this so we won't forget

```
// TODO <Link to issue> We want to deploy external-dns as
// default app to CAPV at some point. Until then, related tests
// are disabled. There's an expiry date for this flag so we notice that we'll have to fix this.
ExternalDnsSupported: time.Now().After(time.Date(2024, 3, 1, 0, 0, 0, 0, time.UTC))
```

common/hello.go
fiunchinho: Yeah, definitely. I'd do it in `clustertest` though.

common/hello.go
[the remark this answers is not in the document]
fiunchinho: You are right! My bad. Thanks

common/hello.go
fiunchinho: Maybe @mcharriere can confirm, but I was under the impression that the default backend will return 404.

common/hello.go
fiunchinho: Not yet, I want to create tickets for all the workarounds documented here after this is merged.

common/hello.go
fiunchinho: But why? I need a string, this is returning `interface{}`.

providers/capv/standard/capv_test.go
fiunchinho: After asking around, support for `external-dns` is part of a big story in provider alignments around DNS. I don't think we have a date for that, and I don't think we should have a deadline for it. I think linking to an issue definitely makes sense.

common/hello.go
fiunchinho: https://github.com/giantswarm/giantswarm/issues/28065

common/hello.go
fiunchinho: https://github.com/giantswarm/giantswarm/issues/28067

common/hello.go
mcharriere: yes, I can confirm, `default-backend` always returns 404.

providers/capz/standard/capz_test.go
AverageMarcus: It is merged but I imagine a new release needs generating first, yeah? Should we get that done now and remove this config flag?

common/hello.go
```diff
@@ -0,0 +1,193 @@
...
+	app.Spec.Namespace = namespace
+	err = managementClusterKubeClient.Create(ctx, app)
+	Expect(err).ShouldNot(HaveOccurred())
+
+	return app, configMap
+}
+
+func getWorkloadClusterDnsZone(ctx context.Context) string {
+	clusterValuesConfigMap := &v1.ConfigMap{}
+	err := state.GetFramework().MC().Get(ctx, ctrl.ObjectKey{Namespace: state.GetCluster().Organization.GetNamespace(), Name: fmt.Sprintf("%s-cluster-values", state.GetCluster().Name)}, clusterValuesConfigMap)
```
AverageMarcus, line 181: Why aren't we using `GetHelmValues` here? 

E.g. 

```go
values = &application.DefaultAppsValues{}
err := state.GetFramework().MC().GetHelmValues(state.GetCluster().Name, state.GetCluster().GetNamespace(), values)
Expect(err).NotTo(HaveOccurred())

if values.BaseDomain == "" {
	Fail("baseDomain field missing from cluster-values configmap")
}

return values.BaseDomain
```

AverageMarcus: `GetHelmValues` calls `GetReleaseValues` on the `helmClient` so should return back the final values used with the chart after all the config layering and defaulting.

common/hello.go
AndiDog: I'd still suggest you check the actual content. Only then, you don't have any assumptions but know for sure you're testing the right thing.

providers/capz/standard/capz_test.go
fiunchinho: The config flag will still be needed, just enabling it `ExternalDnsSupported: true`. We still need to disable it for the non-cloud providers.

common/hello.go
fiunchinho: Done

common/hello.go
fiunchinho: Changed it to use `GetHelmValues`

fiunchinho:
`capv` tests failed, I don't think it's related to this PR

AverageMarcus (approved):

PR #77, opened 2023-08-31
Updated readme with more up-to-date info

AverageMarcus:
### What this PR does

Updated the readme with up-to-date info and added section about running in CI.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,21 @@ Running with Docker:
 docker run --rm -it -v /path/to/kubeconfig.yaml:/kubeconfig.yaml -e E2E_KUBECONFIG=/kubeconfig.yaml quay.io/giantswarm/cluster-test-suites ./
 ```
 
+### ⚙️ Running Tests in CI
+
+These tests are configures to run in our Tekton pipelines with our [cluster-test-suites pipeline](https://github.com/giantswarm/tekton-resources/blob/main/tekton-resources/pipelines/cluster-test-suites.yaml). This pipeline can be triggered on appropriate repos by using the `/run cluster-test-suites` comment trigger.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -95,6 +95,21 @@ Running with Docker:
...
 
+### ⚙️ Running Tests in CI
+
+These tests are configures to run in our Tekton pipelines with our [cluster-test-suites pipeline](https://github.com/giantswarm/tekton-resources/blob/main/tekton-resources/pipelines/cluster-test-suites.yaml). This pipeline can be triggered on appropriate repos by using the `/run cluster-test-suites` comment trigger.
+
+Note: Test suites are configured to run in parallel using a [Matrix](https://tekton.dev/docs/pipelines/matrix/) in the Tekton Pipeline. Once all suites are complete the results of each will be collected and presented in a final Pipeline Task to show the results of all suites.
+
+When the pipeline runs against PRs to the `cluster-test-suites` repo the following is of note:
+* The pipeline will use a container image built from the changes in the PR. The image building and publishing is still handled by CircleCI so the pipeline has a step at the start to wait for this image to be available.
+* The `upgrade` tests won't run as there's no information as to what versions to upgrade from/to.
```
yulianedyalkova, line 106: Can't we do latest released to latest state of main branch? Otherwise adding more upgrade tests for example won't get any validation before getting merged.

yulianedyalkova (approved):

README.md
AverageMarcus: No because there's no expectation that what's in `main` branch is functionally working until a release is cut.

yulianedyalkova: Why not? We're enforcing the tests on each PR and we don't push straight to main so theoretically main should always have green tests, no?

AverageMarcus: I suppose that is true. I guess its just never been agreed that `main` is stable. We could bring it up with the other teams and get their thoughts?

Besides that though, we'd need to implement an extra lookup for the latest commit in main to be able to get the version to upgrade to and then have that passed through as the environment variable to the next step. It's all doable, just adds some more complexity so depends on how much value it adds.

PR #80, opened 2023-08-31
Use cluster-aws defaults

fiunchinho:
### What this PR does
Towards https://github.com/giantswarm/giantswarm/issues/27925

Rely more on `cluster-aws` defaults. This would allow us to always use the right instance type, which we should have reserved on AWS to save money.

### Checklist

### Trigger e2e tests

/run cluster-test-suites


AndiDog (approved):

fiunchinho:
Error in test is unrealted (is about CAPZ), I'm merging this.

PR #107, opened 2023-10-06
Make use of GetExpectedControlPlaneReplicas helper function

AverageMarcus:
### What this PR does

Changes the control plane node tests to make use of the new `GetExpectedControlPlaneReplicas` helper function instead of trying to get the expected number of replicas from the Helm values as this is no longer possible with some providers.

Fixes: https://github.com/giantswarm/giantswarm/issues/28262

### Checklist

### Trigger e2e tests

/run cluster-test-suites


yulianedyalkova (approved):

AverageMarcus:
CAPV seems to be having trouble provisioning clusters right now but is unrelated to this change so I'm going to skip CI as all the other providers passed.

PR #131, opened 2023-11-09
Add autoscaling test for CAPA

njuettner:
### What this PR does

This adds a new common task to scale the test cluster (standard) by creating anti-affinity pods which scales out the cluster by 1 additional node.

The settings for the min and max size of the node pool has been adjusted (2,2 -> 1,2) allowing the cluster-autoscaler creating one more instance.

Issue: https://github.com/giantswarm/roadmap/issues/2772

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/internal/common/scale.go
+++ b/internal/common/scale.go
@@ -9,14 +9,12 @@ import (
 	"github.com/giantswarm/apiextensions-application/api/v1alpha1"
 	"github.com/giantswarm/clustertest/pkg/application"
 	"github.com/giantswarm/clustertest/pkg/client"
-	"github.com/giantswarm/clustertest/pkg/logger"
 	"github.com/giantswarm/clustertest/pkg/wait"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	v1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/types"
 	cr "sigs.k8s.io/controller-runtime/pkg/client"
-	ctrl "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/giantswarm/cluster-test-suites/internal/state"
 )
@@ -73,20 +71,9 @@ func runScale(autoScalingSupported bool) {
 					return false, err
 				}
 
-				now := time.Now()
-				patchedApp := helloApplication.DeepCopy()
-				labels := patchedApp.GetLabels()
-				labels["update"] = fmt.Sprintf("%d", now.Unix())
-				patchedApp.SetLabels(labels)
-
-				err = managementClusterKubeClient.Patch(ctx, patchedApp, ctrl.MergeFrom(helloApplication))
-				if err != nil {
-					return false, err
-				}
-
 				return wait.IsAppDeployed(state.GetContext(), state.GetFramework().MC(), helloApp.InstallName, helloApp.GetNamespace())()
 			}).
-				WithTimeout(15 * time.Minute).
+				WithTimeout(5 * time.Minute).
 				WithPolling(5 * time.Second).
 				Should(BeTrue())
 		})
@@ -99,7 +86,6 @@ func runScale(autoScalingSupported bool) {
 			ctx := context.Background()
 
 			Eventually(func() (string, error) {
-				logger.Log("Trying to get number of nodes from %s", state.GetCluster().Name)
 				helloDeployment := &v1.Deployment{}
 
 				err := wcClient.Get(ctx, cr.ObjectKey{
@@ -111,7 +97,7 @@ func runScale(autoScalingSupported bool) {
 				}
 
 				return strconv.Itoa(int(helloDeployment.Status.ReadyReplicas)), nil
-			}, "30m", "5s").Should(Equal(helloAppValues["ReplicaCount"]))
+			}, "15m", "5s").Should(Equal(helloAppValues["ReplicaCount"]))
 		})
 
 		AfterEach(func() {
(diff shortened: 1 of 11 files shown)


internal/common/scale.go
```diff
@@ -0,0 +1,126 @@
...
+				now := time.Now()
+				patchedApp := helloApplication.DeepCopy()
+				labels := patchedApp.GetLabels()
+				labels["update"] = fmt.Sprintf("%d", now.Unix())
+				patchedApp.SetLabels(labels)
+
+				err = managementClusterKubeClient.Patch(ctx, patchedApp, ctrl.MergeFrom(helloApplication))
+				if err != nil {
+					return false, err
+				}
```
fiunchinho, line 85: I don't think this is needed here. This was used on the other test because it required ingress to be ready, but if we can disable creating `ingress` resources when deploying the app, then the `hello-world` chart should just work the first time.

internal/common/scale.go
```diff
@@ -0,0 +1,126 @@
...
+
+				return wait.IsAppDeployed(state.GetContext(), state.GetFramework().MC(), helloApp.InstallName, helloApp.GetNamespace())()
+			}).
+				WithTimeout(15 * time.Minute).
```
fiunchinho, line 89: Does it need that much time?

internal/common/scale.go
```diff
@@ -0,0 +1,126 @@
...
+				err := wcClient.Get(ctx, cr.ObjectKey{
+					Name:      "scale-hello-world",
+					Namespace: helloApp.InstallNamespace}, helloDeployment)
+
+				if err != nil {
+					return "", err
+				}
+
+				return strconv.Itoa(int(helloDeployment.Status.ReadyReplicas)), nil
+			}, "30m", "5s").Should(Equal(helloAppValues["ReplicaCount"]))
```
fiunchinho, line 114: How much time does it usually take for the autoscaler to do its job? I wouldn't want this waiting 30m every time this test fails.

fiunchinho (commented):
I was wondering if it wouldn't be easier (and specially faster) to just create a `Deployment` with the right `podAntiAffinity`, instead of using the hello world app, which relies on the app platform. What do you think?

fiunchinho:
Any reason not to add it to the `upgrade` suite?

internal/common/scale.go
njuettner: Thanks, I will remove it 👍🏻

internal/common/scale.go
njuettner: Actually it is pretty fast I can set it to 5 minutes

internal/common/scale.go
njuettner: The cluster autoscaler takes a bit of time. I made a mistake I didn't safe the max size changes 🤦🏻

njuettner: I decreased the time by 15 min

njuettner:
@fiunchinho the installation of the app is super quick, so I really would like to keep it. It's also easier to modifying the app values, it's almost everything helm values. 

I don't see a reason not adding it into upgrade suite. I will enable it for upgrade

njuettner:
Launching a new EC2 instance. Status Reason: There is no Spot capacity available that matches your request. Launching EC2 instance failed. | At 2023-11-09T18:17:40Z instances were launched to balance instances in zones eu-west-2c eu-west-2b with other zones resulting in more than desired number of instances in the group. At 2023-11-09T18:18:00Z availability zones eu-west-2c eu-west-2b had 2 0 instances respectively. An instance was launched to aid in balancing the group's zones.
-- | --

Currently hard to test AWS doing AWS things, spot instances FTW

njuettner:
Hmm it seems CAPV for some reason has issues, I didn't touch that

```
[BeforeSuite] [FAILED] [1205.880 seconds]
[BeforeSuite] 
/app/providers/capv/standard/capv_suite_test.go:33

  [FAILED] Unexpected error:
      <context.deadlineExceededError>: 
      context deadline exceeded
      {}
  occurred
  In [BeforeSuite] at: /app/providers/capv/standard/capv_suite_test.go:66 
```

njuettner:
Yeah so CAPV is totally broken, the change is fine for CAPA everything is green

erkanerol:
@njuettner CAPV is broken. We are aware of it. You can ignore it.

fiunchinho (approved):

erkanerol:
CAPV tests are supposed to fail in post cluster creation steps with this error now. We will fix it in default-apps-vsphere level.

```
Common tests cert-manager ClusterIssuers [It] cert-manager default ClusterIssuers are present and ready
/app/internal/common/certmanager.go:34

  [FAILED] Timed out after 30.000s.
  Expected success, but got an error:
      <*errors.StatusError | 0xc000363180>: 
      [clusterissuers.cert-manager.io](http://clusterissuers.cert-manager.io/) "selfsigned-giantswarm" not found
      {
          ErrStatus: {
              TypeMeta: {Kind: "Status", APIVersion: "v1"},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "[clusterissuers.cert-manager.io](http://clusterissuers.cert-manager.io/) \"selfsigned-giantswarm\" not found",
              Reason: "NotFound",
              Details: {
                  Name: "selfsigned-giantswarm",
                  Group: "[cert-manager.io](http://cert-manager.io/)",
                  Kind: "clusterissuers",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 404,
          },
      }
  In [It] at: /app/internal/common/certmanager.go:40 @ 11/13/23 09:55:15.344
```

erkanerol:
CAPV failed where it was supposed to fail. I don't know why CAPA failed.

njuettner:
@erkanerol ok thank you again, CAPA did fail in the last one due to spot issues.

I'll merge the PR.

PR #133, opened 2023-11-13
Release v1.16.2

taylorbot:


erkanerol:
There was no available IP in  `gcapeverde` (CAPV test MC). I cleaned it not rerunning now. This is supposed to pass cluster creation but to fail in cert-manager tests.

/run cluster-test-suites

erkanerol:
CAPV tests failed where it is supposed to fail. We will fix it in default-apps-vsphere level and release a new version.

PR #137, opened 2023-11-20
Increase timeout for default-apps wait

erkanerol:
app-operator sometimes behaves slowly and doesn't reconcile the default-apps immediately. It can take a few minutes in some cases.

Failure example (Re-run worked): https://tekton.ei9o7.k8s.gorilla.eu-central-1.aws.gigantic.io/#/namespaces/tekton-pipelines/pipelineruns/pr-default-apps-vsphere-170-cluster-test-suites98fgh?pipelineTask=run-tests&step=run-tests&taskRunName=pr-default-apps-vsp2b7622e4aa140545bb755731d9bb9df7-run-tests-0

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Increase timeout for default-apps-$provider app.
+
 ## [1.17.0] - 2023-11-15
 
 ### Added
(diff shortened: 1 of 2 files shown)


erkanerol:
@AverageMarcus when you have multiple clusters and apply multiple operations at the same time, it takes some time. If you want, I can observe it for a while.

AverageMarcus:
I'd be interested to know why we only see this with CAPV. The same doesn't seem to occur for CAPA and CAPZ based on the timings I took a little while ago. 🤔 

I'd also like to know if there's something we can do to improve the app-operator to more quickly handle this (it's not really doing anything computationally hard so shouldn't be a slow process) which would then also improve the product for our customers too 😄

erkanerol:
I will observe this issue more. Closing this.

PR #138, opened 2023-11-23
Re-enabled CAPVCD tests

AverageMarcus:
### What this PR does

Re-enabled running the CAPVCD tests now that the new MC has been set up.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Re-enabled CAPVCD tests
+
 ## [1.17.0] - 2023-11-15
 
 ### Added
--- a/providers/capvcd/standard/capvcd_test.go
+++ b/providers/capvcd/standard/capvcd_test.go
@@ -6,7 +6,7 @@ import (
 	"github.com/giantswarm/cluster-test-suites/internal/common"
 )
 
-var _ = PDescribe("Common tests", func() {
+var _ = Describe("Common tests", func() {
 	common.Run(&common.TestConfig{
 		BastionSupported: true,
 		// Disabled until https://github.com/giantswarm/roadmap/issues/1037


ericgraf (approved):

AverageMarcus:
@glitchcrab 
```
[BeforeSuite] [FAILED] [0.034 seconds]
[BeforeSuite] 
/app/providers/capvcd/standard/capvcd_suite_test.go:29

  [FAILED] Unexpected error:
      <*url.Error | 0xc000888060>: 
      Get "https://178.170.32.23:6443/api?timeout=32s": Service Unavailable
      {
          Op: "Get",
          URL: "https://178.170.32.23:6443/api?timeout=32s",
          Err: <*errors.errorString | 0xc00043a000>{
              s: "Service Unavailable",
          },
      }
  occurred
  In [BeforeSuite] at: /app/providers/capvcd/standard/capvcd_suite_test.go:61 @ 11/23/23 16:30:11.554
```

AverageMarcus:
Looks like the kubeconfig secret wasn't correctly updated. The new MC wasn't named `capvcd` so it wasn't being found. I've now fixed that so lets try again...

/run cluster-test-suites

AverageMarcus:
There's still a few issues:

The cluster app ends up with the following:
> parsing error: failed to parse `vcd-credentials` in `org-giantswarm`, logs: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type map[string]interface {} 

Looking at that secret I don't actually think its needed in CAPVCD (like it is in CAPV) so I tried removing it and then the cluster app ended up with the following:

```
values don't meet the specifications of the schema(s) in the following chart(s):
cluster-cloud-director:
- connectivity: Additional property bastion is not allowed
- controlPlane.oidc: clientId is required
- controlPlane.oidc: groupsClaim is required
- controlPlane.oidc: issuerUrl is required
- controlPlane.oidc: usernameClaim is required
```

I suspect the values files need updating with the latest changes needed for cluster-cloud-director.

@giantswarm/team-rocket is one of you able to take over this PR and get the fixes added so we can get E2E running for CAPVCD again please? 😄

AverageMarcus:
I just want to test webhook are still working after updating the configuration: 

/run cluster-test-suites

glitchcrab:
ref https://github.com/giantswarm/giantswarm/issues/27838

AverageMarcus:
I'm re-triggering as I want to check that an update I made to a Task worked...

/run cluster-test-suites

providers/capvcd/standard/test_data/cluster_values.yaml
```diff
@@ -3,38 +3,33 @@ controlPlane:
...
   nodeClasses:
     default:
       catalog: giantswarm
       sizingPolicy: m1.large
-      template: ubuntu-2004-kube-v1.22.9
+      template: ubuntu-2004-kube-v1.24.10
   userContext:
     secretRef:
       secretName: vcd-credentials
+  vmNamingTemplate: neoedge{{.machine.Name | sha256sum | trunc 4}}
```
AverageMarcus, line 30: This file is handled as a Go template file, you'll need to escape these if you want to have them passed through to the actual Helm values.

On a side note, there's a LOT of values in here. WAY more than I'd expect a custom to need to provide. Are they all required? Where possible we should be relying on the chart default values otherwise this is going to be quite brittle as things get updated.

glitchcrab: yes these are all required, there are no defaults between customer vsphere environments so we need customers to provide them. unfortunately this set of values is the bare minimum required.

glitchcrab:
weird all providers failed - might have been a network hiccup

glitchcrab:
merging this in as CAPVCD tests passed

erkanerol:
Do we need a release to make this PR effective in other repos like cluster-cloud-director and default-apps-cloud-director? I think so.

PR #155, opened 2023-12-04
Fix observability-bundle failing tests

QuantumEnigmaa:
### What this PR does

This PR increases the timeout period for observability-bundle apps from 15 to 25min. As some tests run by the CI are failing whereas all those run manually succeeded, I guess this is mainly caused by the short timeout period which doesn't give enough time for the apps to deploy. 

If this doesn't work, I'll try to add more log outputs for the tests to have a description of the apps' events.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Increased timeout for observability-bundle apps.
+
 ## [1.19.2] - 2023-12-04
 
 ### Fixed
--- a/internal/common/apps.go
+++ b/internal/common/apps.go
@@ -64,7 +64,7 @@ func runApps() {
 			}
 
 			Eventually(wait.IsAllAppDeployed(state.GetContext(), state.GetFramework().MC(), appNamespacedNames)).
-				WithTimeout(15 * time.Minute).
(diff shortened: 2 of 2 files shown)


AverageMarcus:
@QuantumEnigmaa Have you seen examples of it taking more than 15 minutes for the apps to be installed? I'm always resistant to just bumping the timeout as a first attempt as it will cause a delay in reporting back to people waiting for their PRs.

QuantumEnigmaa:
> Have you seen examples of it taking more than 15 minutes for the apps to be installed?

I have seen it with some WCs creation but not as a regular thing. And for e2e testing I was never able to find the root cause before the end of the tests. :/ 

> I'm always resistant to just bumping the timeout as a first attempt as it will cause a delay in reporting back to people waiting for their PRs.

Yeah I understand that. I can first add the additional log outputs and go back to this PR if it confirms that an added delay is the way to go.

AverageMarcus:
Lets give this PR a try for now. We can see if it improves the daily runs or not. :)

AverageMarcus (approved):

AverageMarcus:
I think there's an issue with cert-manager (maybe rate-limiting from lets encrypt) as I'm seeing the same failing tests on other PRs. I'm going to add the skip label as this Pr is good to go.

AverageMarcus:
@QuantumEnigmaa Feel free to merge and put out a new patch release

PR #162, opened 2023-12-12
eks-update-global-values

calvix:
update values to match new scheme https://github.com/giantswarm/cluster-eks/pull/50

### DIFF
--- a/providers/eks/standard/test_data/cluster_values.yaml
+++ b/providers/eks/standard/test_data/cluster_values.yaml
@@ -17,6 +17,7 @@ global:
   # defaulted in the chart template.
   # @TODO: https://github.com/giantswarm/giantswarm/issues/28063
   nodePools:
+    # We are using a name with 10 chars which is the max number of characters allowed by our kyverno policies.
     nodepool-0:
       instanceType: r6i.xlarge
       maxSize: 3
--- a/providers/eks/standard/test_data/cluster_values.yaml
+++ b/providers/eks/standard/test_data/cluster_values.yaml
@@ -1,23 +1,24 @@
-metadata:
-  name: "{{ .ClusterName }}"
-  description: "E2E Test cluster"
-  organization: "{{ .Organization }}"
+global:
+  metadata:
+    name: "{{ .ClusterName }}"
+    description: "E2E Test cluster"
+    organization: "{{ .Organization }}"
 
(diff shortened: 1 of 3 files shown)


calvix:
EKS test passed, not sure what is the rest but this changes only values for EKS so it should not really matter for the rest.

and its kinda needed to merge to avoid errors for e2e for ESK as the chart already has the new scheme

providers/eks/upgrade/test_data/cluster_values.yaml
```diff
@@ -1,24 +1,24 @@
...
-    instanceType: r6i.xlarge
-    maxSize: 3
-    minSize: 3
-    rootVolumeSizeGB: 25
+  # We need to pass the node pools otherwise the test called "has all the worker nodes running" will fail because it
+  # expects to find the number of worker nodes in the helm values, but the value is not there if don't pass it, as it's
+  # defaulted in the chart template.
+  # @TODO: https://github.com/giantswarm/giantswarm/issues/28063
+  nodePools:
+    nodepool-0:
```
fiunchinho, line 20:
```suggestion
    # We are using a name with 10 chars which is the max number of characters allowed by our kyverno policies.
    nodepool-0:
```

providers/eks/standard/test_data/cluster_values.yaml
```diff
@@ -1,24 +1,24 @@
...
-    instanceType: r6i.xlarge
-    maxSize: 3
-    minSize: 3
-    rootVolumeSizeGB: 25
+  # We need to pass the node pools otherwise the test called "has all the worker nodes running" will fail because it
+  # expects to find the number of worker nodes in the helm values, but the value is not there if don't pass it, as it's
+  # defaulted in the chart template.
+  # @TODO: https://github.com/giantswarm/giantswarm/issues/28063
+  nodePools:
+    nodepool-0:
```
fiunchinho, line 20:
```suggestion
    # We are using a name with 10 chars which is the max number of characters allowed by our kyverno policies.
    nodepool-0:
```

calvix:
not sure what is happening with all the test but the EKS test itself pass it fails at the end when its trying to remove the org, this was not the case yesterday

calvix:
yay green!

njuettner (approved):

PR #164, opened 2023-12-13
capa-disable-bastions

calvix:
towards https://github.com/giantswarm/giantswarm/issues/29254
disable bastion test for CAPA as we want to remove it in this PR https://github.com/giantswarm/cluster-aws/pull/455


AndiDog (approved):

calvix:
both CAPA test  are successful multiple times so merging anyway

PR #169, opened 2023-12-20
Reduce the number of CP nodes from 3 to 1 in CAPVCD tests

erkanerol:
### What this PR does

We hit timeouts while waiting for nodes since CAPVCD is super slow in node provisioning. We think it is OK to have 1 CP in WC tests.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Reduce the number of CP nodes from 3 to 1 in CAPVCD tests in order to avoid timeouts.
+
 ## [1.20.3] - 2023-12-19
 
 ### Added
--- a/providers/capvcd/standard/test_data/cluster_values.yaml
+++ b/providers/capvcd/standard/test_data/cluster_values.yaml
@@ -1,6 +1,8 @@
 baseDomain: test.gigantic.io
 controlPlane:
   catalog: giantswarm
+  # HA is disabled temporarily due to
+  # https://github.com/giantswarm/giantswarm/issues/29353
   replicas: 1
   sizingPolicy: m1.large
   template: ubuntu-2004-kube-v1.24.10
--- a/providers/capvcd/standard/test_data/cluster_values.yaml
+++ b/providers/capvcd/standard/test_data/cluster_values.yaml
@@ -1,7 +1,7 @@
 baseDomain: test.gigantic.io
 controlPlane:
   catalog: giantswarm
-  replicas: 3
+  replicas: 1
   sizingPolicy: m1.large
   template: ubuntu-2004-kube-v1.24.10
   oidc:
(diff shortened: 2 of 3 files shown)


AverageMarcus (commented):
I thought we weren't going to support single control plane with CAPI. Has that changed?

erkanerol:
@AverageMarcus I am not sure about it. I got your point. It doesn't make sense to test the single-CP case if we mandate multi-CP in production. On the other hand, we have other kinds of tests like MC creation and we can catch multi-CP issues there if they exist.

We need to wait for a lot of time at the moment because multi-nodes and a single CP is enough to test the basic functionality.

erkanerol:
Node we also use a single CP for CAPV. https://github.com/giantswarm/cluster-test-suites/blob/main/providers/capv/standard/test_data/cluster_values.yaml#L9

glitchcrab (approved):
👌

erkanerol:
The second run failed for EKS at an unrelated point. I want to merge it if @AverageMarcus you are OK.

AverageMarcus:
So are we going to support single control plane replicas with our customers then? If so, then I guess this is fine to merge in but I wasn't aware of that change in decision so I suspect others aren't also. Might be worth a note in kaas-sync or something like that.

erkanerol:
> So are we going to support single-control plane replicas with our customers then? 

I don't know. I haven't said so. 

My point is we are OK to test with only one CP here to avoid timeouts by accepting the loss of capability of catching bugs in the multi-CP scenario. We will rely on other tests for that case.

AverageMarcus:
We don't have any other WC tests that test multi-cp capabilities if you're removing it from here. The MC tests don't cover enough. The whole point of these WC tests is to cover the basic use cases that our customers are going to be doing.

erkanerol:
> We don't have any other WC tests that test multi-cp capabilities if you're removing it from here. The MC tests don't cover enough. The whole point of these WC tests is to cover the basic use cases that our customers are going to be doing.

I understand Marcus' point here. 

@gawertm @glitchcrab  If we really want to go with 3 CP, we should 1) increase timeouts in more places 2) investigate&improve CAPVCD manager. It is incredibly slow.

I don't think I will have time for this in the mid-term. The options are

- We can open an issue and live with flaky tests.
- We can prioritize this issue and leave other stuff for a while.
- Someone else can take this.
- We can live with 1 CP test for a while and then fix it.

I prefer the last one but it is up to you.

AverageMarcus:
Just to clarify, I wont block this change if you think it's best and you're ok living with the not-quite-how-customers-run-things approach. I've just here to point out that the tests might miss some things because of it. 😄 

If it was up to me, speeding things up in CAPVCD would be my preference but I know (from experience) that it might not be possible or easy to do so.

erkanerol:
I created a PM regarding slowness of CAPVCD https://github.com/giantswarm/giantswarm/issues/29353 FYI.

AverageMarcus:
Thanks for the PM @erkanerol 👍  Can we add a comment linking to that issue alongside the change in the yaml and then I'm happy for this change to go through to "fix" things for now. 😄

erkanerol:
@AverageMarcus  I updated yaml files and added the PM reference. I also add a note to the PM to not forget to revert this change.

AverageMarcus (approved):

PR #175, opened 2024-01-11
Allow basic test to retry on flake

AverageMarcus:
### What this PR does

Allows tests to retry if they fail due to, for example, flakey network.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/internal/common/basic.go
+++ b/internal/common/basic.go
@@ -30,11 +30,11 @@ func runBasic() {
 			}
 		})
 
-		It("should be able to connect to MC cluster", func() {
+		It("should be able to connect to MC cluster", FlakeAttempts(3), func() {
 			Expect(state.GetFramework().MC().CheckConnection()).To(Succeed())
 		})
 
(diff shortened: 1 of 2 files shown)


internal/common/basic.go
```diff
@@ -30,11 +30,11 @@ func runBasic() {
 			}
 		})
 
-		It("should be able to connect to MC cluster", func() {
+		It("should be able to connect to MC cluster", FlakeAttempts(3), func() {
```
yulianedyalkova, line 33: Oh nice, I didn't even know this exists in Ginkgo!

yulianedyalkova (approved):

internal/common/basic.go
AverageMarcus: Yeah it's pretty nice. Rather than having a full retry of the whole suite 😄 

I was reluctant to use it initially as ideally don't want flakes to even exist in the tests but networking is what it is 😅

PR #178, opened 2024-01-11
Fix ClusterIssuer Eventually logic

AverageMarcus:
### What this PR does

Fix the function used in the `Eventually` call to actually return a function rather than an error.

### Checklist

### Trigger e2e tests

/run cluster-test-suites


mnitchev (approved):

AverageMarcus:
CAPVCD is currently having issues on the MC so that failing test can be ignored for now. Adding the skip label as all other tests have passed.

PR #181, opened 2024-01-12
Refactor and consolidate suite setup code; introduce AfterSuite for better cleanup

AverageMarcus:
### What this PR does

Introduces a new shared module `suite` that provides a common way of handling the test suite setup and clean up rather than having duplicated code.

The new module introduces an `AfterSuite` to replace the previously used `DeferCleanup` that handles WC cluster deletion.  This works towards https://github.com/giantswarm/giantswarm/issues/29377 and should ensure that failures during the `BeforeSuite` still get cleaned up if possible.

As part of refactoring `standup` to use the new code I noticed that EKS support was missing so that has also been added with the refactor.

### Checklist

### Trigger e2e tests

/run cluster-test-suites


AverageMarcus:
Only test failure is a DNS flake so I'm adding the skip label.

internal/suite/setup.go
```diff
@@ -0,0 +1,96 @@
...
+
+// ClusterBuilder is an interface that provides a function for building provider-specific Cluster apps
+type ClusterBuilder interface {
+	NewClusterApp(clusterName string, orgName string, clusterValuesFile string, defaultAppsValuesFile string) *application.Cluster
+}
+
+// Setup handles the creation of the BeforeSuite and AfterSuite handlers. This covers the creations and cleanup of the test cluster.
+// `clusterReadyFns` can be provided if the cluster requires custom checks for cluster-ready status. If not provided the cluster will
+// be checked for at least a single control plane node being marked as ready.
+func Setup(isUpgrade bool, kubeContext string, clusterBuilder ClusterBuilder, clusterReadyFns ...func(client *client.Client)) {
```
yulianedyalkova, line 31: Just a thought: you can just pass the NewClusterApp function to Setup instead of the interface so you don't have to create so many new types and then pass empty objects around just to use the one function (unless we see the interface and types growing in the future).

yulianedyalkova: Or even just pass the provider and let Setup decide which library to call out to.

AverageMarcus: Wont that lead to cyclical imports?

AverageMarcus: Actually, it wont thinking about it. But I'm not sure the provider would work as we need different logic for CAPA with private networking, for example.

yulianedyalkova: > But I'm not sure the provider would work as we need different logic for CAPA with private networking, for example.

We can then use the KubeContext buuut then we have to store our own mapping so I don't like it anymore :D

> That was my thinking. Not sure what that might be yet. Maybe nothing. What do you think?

I can't currently think of anything either but who knows. Happy to keep it this way as we have it implemented already then :)

yulianedyalkova (approved):

internal/suite/setup.go
AverageMarcus: We could always refactor it to just pass a function at a later date if we decide thats enough. :)

PR #223, opened 2024-02-21
Add teleport connectivity test

tuladhar:
Towards https://github.com/giantswarm/giantswarm/issues/29741

### What this PR does
- Adds teleport connectivity test
- Go 1.21 - Teleport API package requires Go >1.21
```
/Users/puru/go/pkg/mod/github.com/gravitational/teleport/api@v0.0.0-20240220132348-146d3194f72a/utils/aws/identifiers.go:21:2: package slices is not in GOROOT (/opt/homebrew/Cellar/go/1.20.5/libexec/src/slices)
note: imported by a module that requires go 1.21
```

### Merge checklist
- [x] https://github.com/giantswarm/cluster-test-suites/pull/224
- [x] Document teleport env var in README
- [x] Deploy tbot in tekton cluster (dev, prod) to generate secret containing Teleport identity file token. https://github.com/giantswarm/workload-clusters-fleet/pull/853 https://github.com/giantswarm/workload-clusters-fleet/pull/859
- [x] Update CICD pipeline to mount teleport identity file from secret [here](https://github.com/giantswarm/tekton-resources/blob/1e3521aaa5d09a214b92f81c7add8fdde19015b0/tekton-resources/tasks/run-wc-e2e-tests.yaml#L70) - https://github.com/giantswarm/tekton-resources/pull/267
- [x] Rename `identity-output` and add label to the secret if possible.

### Checklist

### Trigger e2e tests

/run cluster-test-suites


ssyno (approved):

internal/common/teleport.go
```diff
@@ -0,0 +1,62 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func runTeleport(teleportSupported bool) {
+	Context("teleport", func() {
+		var teleportClient *tc.Client
+
+		isClusterRegistered := func(clusterName string) (bool, error) {
+			clusters, err := teleportClient.GetKubernetesServers(context.Background())
```
tuladhar, line 24: turns out `GetKubernetesCluster` doesn't return any registered cluster, so had to use `GetKubernetesServers`

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.20 AS build-tests
```
AverageMarcus, line 1: No! Please don't bump the Go version!

go.mod
```diff
@@ -1,10 +1,11 @@
 module github.com/giantswarm/cluster-test-suites
 
-go 1.20
```
AverageMarcus, line 3: Please don't change the Go version.
Is there something specifically in Go 1.21 you require for this change? If not, please keep this separate.

internal/common/teleport.go
```diff
@@ -0,0 +1,62 @@
...
+			return false, nil
+		}
+
+		BeforeEach(func() {
+			if strings.TrimSpace(os.Getenv("TELEPORT_IDENTITY_FILE")) == "" {
+				Skip("TELEPORT_IDENTITY_FILE env var not set, skipping teleport test")
+				return
+			}
+			var err error
+			teleportClient, err = helper.NewTeleportClient(context.Background(), os.Getenv("TELEPORT_IDENTITY_FILE"))
```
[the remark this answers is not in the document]
AverageMarcus: I have the same question, I don't see any documentation updates related to this. This isn't set in the CI/CD pipelines.

internal/common/teleport.go
```diff
@@ -0,0 +1,62 @@
...
+			Expect(err).To(BeNil())
+		})
+
+		It("cluster is registered", func() {
+			if !teleportSupported {
+				Skip("Teleport is not supported.")
+			}
```
AverageMarcus, line 51: Move this into the `BeforeEach`

internal/helper/teleport.go
```diff
@@ -0,0 +1,29 @@
+package helper
+
+import (
+	"context"
+
+	tc "github.com/gravitational/teleport/api/client"
+)
+
+var NewTeleportClient = func(ctx context.Context, identityFilePath string) (*tc.Client, error) {
```
AverageMarcus, line 9: Can we please follow the same syntax used by the rest of the project and have this as a name function instead of a variable.

Also, it might make sense to have this as it's own module and use the standard `New` function to create a client.

AverageMarcus (changes requested):
Some comments.

Can we please make sure the new env var is documented so others know how to run this. This will also require changes to our CI/CD pipelines in Tekton which should be implemented before this is merged (so we can have the tests on this PR successfully pass with the Teleport tests enabled)

go.mod
AverageMarcus: > Go 1.21 - Teleport API package requires Go >1.21

Sorry, I missed this. 

Could this please be done as an individual PR so that it can be easily tracked in the git history?

internal/common/teleport.go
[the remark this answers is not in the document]
tuladhar: So, it will be set in CI/CD pipeline, [similar to GITHUB_TOKEN_FILE](https://github.com/giantswarm/tekton-resources/blob/1e3521aaa5d09a214b92f81c7add8fdde19015b0/tekton-resources/tasks/run-wc-e2e-tests.yaml#L69-L70)

However, with Teleport we can't really create long lived token, so we need to deploy an tbot agent in the cluster similar to how we are doing with teleport-operator, that periodically refreshes the token.

go.mod
tuladhar: Yes, good point, will take the Go version change separately.

internal/helper/teleport.go
tuladhar: Thanks! Yeah, making it's own module makes sense.

internal/common/teleport.go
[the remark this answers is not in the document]
AverageMarcus: Can we please tackle that before this PR gets merged in? It's no use merging this in if it's not going to actually be used.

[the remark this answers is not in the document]
tuladhar: Yep, won't be merging this until identity file is sorted out.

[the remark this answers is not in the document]
AverageMarcus: I've added the `do-not-merge/hold` label to be sure :)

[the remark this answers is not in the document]
tuladhar: `teleport-tbot` deployed in `gazelle/cicddev` and identity file sorted.

internal/common/teleport.go
```diff
@@ -0,0 +1,61 @@
...
+	tc "github.com/gravitational/teleport/api/client"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func runTeleport(teleportSupported bool) {
+	Context("teleport", func() {
+		var teleportClient *tc.Client
+
+		isClusterRegistered := func(clusterName string) (bool, error) {
```
AverageMarcus, line 23: What's the point of this variable rather than just inlining the function in the `It` below?

AverageMarcus (changes requested):
Can you please also update the documentation to explain about the new environment variable and how someone might run these tests locally with it set

tuladhar:
> Some comments.
> 
> Can we please make sure the new env var is documented so others know how to run this. This will also require changes to our CI/CD pipelines in Tekton which should be implemented before this is merged (so we can have the tests on this PR successfully pass with the Teleport tests enabled)

@AverageMarcus I've documented the teleport env var, along with changes to CI/CD pipeline are done. The PR successfully passed the Teleport test.

internal/common/teleport.go
tuladhar: Hmm...I used it based on similar approach in `dns.go`
> https://github.com/giantswarm/cluster-test-suites/blob/d5285e05b9710e45af8f3d40db4374de68b41133/internal/common/dns.go#L24-L33

AverageMarcus: But your function is only used in a single place. It's not shared across tests so no need to have it as a variable unless there's something I'm missing.

AverageMarcus: Also, the way that Ginkgo runs code can be a little surprising sometimes so where possible you should avoid running login outside of an `It` or the `Before/After` hooks.

tuladhar: Okay, gotcha.

tuladhar: Addressed in commit https://github.com/giantswarm/cluster-test-suites/pull/223/commits/262d74f4fabca2a41614ae82d2bf58a386641c0e

AverageMarcus (approved):

tuladhar:
✅ Teleport test passed, but CAPZ and CAPVCD failed which is unrelated to this changes.
```
Common tests teleport cluster is registered
/app/internal/common/teleport.go:36
  {"level":"info","ts":"2024-02-29T11:28:25Z","msg":"cluster registered KubeServer(Name=grizzly-t-2hln6qtyvna4c5kf55, Version=14.1.3, Hostname=remote.kube.proxy.teleport.cluster.local, HostID=2eb79f17-80f2-4f23-9f44-41d28c1a33e5, Cluster=KubernetesCluster(Name=grizzly-t-2hln6qtyvna4c5kf55, Labels=map[]))"}
• [0.400 seconds]
```

Merging with `skip/ci`

PR #236, opened 2024-03-11
Add a new test that checks if key workload cluster metrics exist in p…

whites11:
…rometheus.

### Checklist

### Trigger e2e tests

/run cluster-test-suites


internal/common/metrics.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/giantswarm/clustertest/pkg/logger"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/giantswarm/cluster-test-suites/internal/helper"
+	"github.com/giantswarm/cluster-test-suites/internal/state"
+)
+
+func runMetrics() {
+	Context("default apps", func() {
```
AverageMarcus, line 19:
```suggestion
	Context("metrics", func() {
```

internal/common/metrics.go
```diff
@@ -0,0 +1,129 @@
...
+		var mcClient *client.Client
+
+		BeforeEach(func() {
+			mcClient = state.GetFramework().MC()
+		})
+
+		// List of metrics that must be present.
+		metrics := []string{
```
AverageMarcus, line 27: Might be best to move this into either the `BeforeEach` or the `It` as otherwise the ordering of code execution can be a bit unexpected if tests (in the future) make changes to this slice.

internal/helper/podrunner.go
```diff
@@ -0,0 +1,64 @@
...
+	tty := true
+
+	restCfg, err := controllerruntime.GetConfig()
+	if err != nil {
+		return "", "", microerror.Mask(err)
+	}
+	coreClient, err := kubernetes.NewForConfig(restCfg)
+	if err != nil {
+		return "", "", microerror.Mask(err)
+	}
```
AverageMarcus, line 31: We already have Kubernetes clients for both the MC and the WC that should be used rather than creating new clients:
E.g.
```
state.GetFramework().MC()
```

internal/common/metrics.go
whites11: done

AverageMarcus (approved):

PR #249, opened 2024-03-21
Increase cluster creation timeout to 30 minutes

glitchcrab:
### What this PR does

Increases the cluster creation timeout to try and avoid test flakes caused by slow infra.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Increase cluster creation timeout to 30 minutes.
+
 ## [1.32.0] - 2024-03-20
 
 ### Changed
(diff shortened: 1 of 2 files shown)


glitchcrab:
I don't know if there's a better way of achieving this, but at the moment we see a fair amount of test failures on VCD when the infra takes longer than 20 minutes to come up. We would like to try and optimise the vcd-controller to improve this but it's not a small task. In the short term, increasing the timeout should help reduce the number of unnecessary failures.

ericgraf (approved):
Crazy to think that it could take up to 30 minutes to bring up a cluster. It is a very complex process with multi master...etc but at somepoint this performance will have to investigated.

glitchcrab:
Yes, I don't like it either to be honest but I regularly see tests fail at the 1200s (20m) mark :disappointed:

glitchcrab:
i'll merge this in - the CAPA tests failed but this PR wouldn't be the cause of that

PR #251, opened 2024-03-26
capvcd 1.25.13

vxav:
### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Increase cluster creation timeout to 30 minutes.
+- Switch CAPVCD values to 1.25.13.
 
 ## [1.32.0] - 2024-03-20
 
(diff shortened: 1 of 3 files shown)


glitchcrab:
do we not want to drop the values here to consume the defaults from the cluster chart?

AverageMarcus (approved):

AverageMarcus:
Also while I think about it, Ubuntu 20.04 is a bit dated now isn't it? Any reason not to use 22.04 until ignition is supported in CAPVCD?

vxav:
If we don't get a decision on the nodeClass thing I'll just set the default values in the less ideal way but it is planned yes.

Regarding 20.04 I did try 22.04 a long time ago and I had lots of issues with cgroups with CAPVCD. However Flatcar is around the corner (PR merged upstream).

AverageMarcus:
> However Flatcar is around the corner (PR merged upstream).

Yay! 

Does that mean we need to add CAPVCD support for capi-image-builder? Or does it use the same as CAPV?

vxav:
We can use the same as CAPV.

PR #252, opened 2024-03-27
Increase cert-manager test timeout to 15 minutes

tuladhar:
Towards https://github.com/giantswarm/giantswarm/issues/30364

### What this PR does
- Increases the timeout when waiting for cert-manager to 15 minutes.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Increase cluster creation timeout to 30 minutes.
 - Switch CAPVCD values to 1.25.13.
+- Increase cert-manager test timeout to 15 minutes.
 
 ## [1.32.0] - 2024-03-20
 
--- a/internal/common/certmanager.go
+++ b/internal/common/certmanager.go
@@ -8,6 +8,7 @@ import (
 	"github.com/giantswarm/cluster-test-suites/internal/state"
 	"github.com/giantswarm/clustertest/pkg/client"
 	"github.com/giantswarm/clustertest/pkg/logger"
+	"github.com/giantswarm/clustertest/pkg/wait"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	cr "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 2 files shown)


ericgraf (changes requested):
This is a big jump in the timeout. Can you explain why we need such a big timeout? 
We need to fail fast to give a better user experience when running tests.

tuladhar:
> This is a big jump in the timeout. Can you explain why we need such a big timeout? We need to fail fast to give a better user experience when running tests.

@ericgraf I agree with fail fast approach but [CAPVCD E2E tests are slow](https://tekton.ci.giantswarm.io/#/namespaces/tekton-pipelines/pipelineruns/pr-default-apps-cloud-director-234-cluster-test-suites6lfwd?pipelineTask=run-tests&taskRunName=pr-default-apps-clo2fb6acfb02d7982a2b5b8d855d84a36a-run-tests-1&step=run-tests) in general, and in cert-manager case the job that installs  `cluster issuer` takes time to install. [See thread](https://gigantic.slack.com/archives/C04KA9DBMB2/p1711454921752519)

The [job that installs the `cluster issuer` retries for 15 minutes](https://github.com/giantswarm/cert-manager-app/blob/ae80875318c544e65fb10ca5def2920cf8216953/helm/cert-manager/charts/cert-manager-giantswarm-clusterissuer/templates/job.yaml#L34) and timeout in test of 2 minutes is not enough for CAPVCD.

For CAPA, the cert-manager test is passing fine, and we haven't experienced this issue.

cc: @ssyno

ssyno (approved):

AverageMarcus:
@tuladhar if the issue is with slowness on CAPVCD (and not other providers) then this test is working and indicating a problem that needs to be fixed. The timeout shouldn't be increased, the issue in CAPVCD needs to be addressed instead. 
What work has been done in CAPVCD to solve this? Is there an open issue for the underlying problem? What is the cause for this only to occur on one provider?

tuladhar:
> @tuladhar if the issue is with slowness on CAPVCD (and not other providers) then this test is working and indicating a problem that needs to be fixed. The timeout shouldn't be increased, the issue in CAPVCD needs to be addressed instead. What work has been done in CAPVCD to solve this? Is there an open issue for the underlying problem? What is the cause for this only to occur on one provider?

Yep, the issue is only CAPVCD and increasing timeout is bad way to fix the underlying issue.

[In this thread](https://gigantic.slack.com/archives/C04KA9DBMB2/p1711454921752519), the team adjusted [VPN DNS setting](https://github.com/giantswarm/vpn/pull/460/files) which seems to improve the E2E tests. Let's see if we hit cert-manager timeout issue again, after these changes.

Meanwhile, closing this issue.

PR #259, opened 2024-04-08
Add support for mimir in the `metrics` test.

whites11:
### What this PR does

Towards:

https://github.com/giantswarm/giantswarm/issues/30424

adds support for mimir to the `metrics` test.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/internal/common/metrics.go
+++ b/internal/common/metrics.go
@@ -7,7 +7,6 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/giantswarm/clustertest/pkg/client"
 	"github.com/giantswarm/clustertest/pkg/logger"
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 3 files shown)


internal/common/metrics.go
```diff
@@ -132,3 +155,120 @@ func checkMetricPresent(mcClient *client.Client, namespace string, podName strin
...
+			return fmt.Errorf("error ensuring test pod is running %s: %s", podName, err)
+		}
+
+		if existing.Status.Phase == corev1.PodRunning {
+			break
+		}
+
+		logger.Log("Waiting for pod %s to be running", podName)
+		time.Sleep(5 * time.Second)
+	}
```
AverageMarcus, line 239: Instead of using these endless `for` loops could we instead please use something like `Eventually` with a timeout so that we don't wait forever if something gets stuck?

Something like... (I haven't tested this)

```
Eventually(func() (bool, error) {
	err = mcClient.Get(context.Background(), client2.ObjectKey{Namespace: ns, Name: podName}, &existing)
	if err != nil {
		return false, fmt.Errorf("error ensuring test pod is running %s: %s", podName, err)
	}

	if existing.Status.Phase == corev1.PodRunning {
		return true, nil
	}

	logger.Log("Waiting for pod %s to be running", podName)
	return false, nil
}).
	WithTimeout(5 * time.Minute).
	WithPolling(5 * time.Second).
	Should(BeTrue())
```


Same with below when deleting.

whites11: done, testing in progress

whites11: seems to work

AverageMarcus (approved):
👨‍🍳 🤏

PR #263, opened 2024-04-16
use default values of cluster-cloud-director chart:v0.50.0

vxav:
### What this PR does

In `cluster-cloud-director` `v0.50.0`, we move to a release model where K8s version and VM template are set in default values to be controlled by Giant Swarm through cluster app versions.

Towards https://github.com/giantswarm/roadmap/issues/3392

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/providers/capvcd/standard/test_data/cluster_values.yaml
+++ b/providers/capvcd/standard/test_data/cluster_values.yaml
@@ -1,6 +1,7 @@
 baseDomain: test.gigantic.io
 controlPlane:
   replicas: 3
+  diskSizeGB: 30
   sizingPolicy: m1.large
   oidc:
     clientId: "dex-k8s-authenticator"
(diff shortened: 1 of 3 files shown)


erkanerol (commented):
LGTM

providers/capvcd/standard/test_data/cluster_values.yaml
```diff
@@ -1,11 +1,8 @@
 baseDomain: test.gigantic.io
 controlPlane:
-  catalog: giantswarm
   replicas: 3
+  diskSizeGB: 30
```
AverageMarcus, line 4: How come this is needed for standard but not upgrade suite?

vxav: It was an oversight on my part, thanks.
Fixed

AverageMarcus: 😕  I still see it on the PR. Is it just me?

AverageMarcus: Oh, you need it on both! Got it!

PR #266, opened 2024-04-18
Check if Cluster is ready and if control plane rolling update has finished

nprokopic:
### What this PR does

- Add common basic test that checks if Cluster Ready condition has Status=True.
- Add upgrade test that checks if control plane rolling update has finished (if it has started).

The second test here is the more important one, as it makes sure that a control plane rolling update has finished before it executes other tests.

Before having this control plane rolling update test, other post-upgrade tests would continue running even though the cluster upgrade has not been completed, so those test would make checks in a cluster that has a mix of old and new nodes.

Now with the new control plane rolling update test at least control plane nodes should be done with a rolling update, so the subsequent checks will be more reliable.

### Checklist

- [x] Update clustertest to a released version (waiting on https://github.com/giantswarm/clustertest/pull/166)

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.21
 
 require (
 	github.com/giantswarm/apiextensions-application v0.6.1
-	github.com/giantswarm/clustertest v0.17.1-0.20240418115808-bea0f65c53bc
+	github.com/giantswarm/clustertest v0.18.0
 	github.com/gravitational/teleport/api v0.0.0-20240221014947-7cec562a5fe2
 	github.com/onsi/ginkgo/v2 v2.17.1
 	github.com/onsi/gomega v1.32.0
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.21
 
 require (
 	github.com/giantswarm/apiextensions-application v0.6.1
-	github.com/giantswarm/clustertest v0.17.1-0.20240418092820-fb25dad2e6c8
+	github.com/giantswarm/clustertest v0.17.1-0.20240418115808-bea0f65c53bc
 	github.com/gravitational/teleport/api v0.0.0-20240221014947-7cec562a5fe2
 	github.com/onsi/ginkgo/v2 v2.17.1
 	github.com/onsi/gomega v1.32.0
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.21
 
 require (
 	github.com/giantswarm/apiextensions-application v0.6.1
-	github.com/giantswarm/clustertest v0.17.0
+	github.com/giantswarm/clustertest v0.17.1-0.20240418092820-fb25dad2e6c8
 	github.com/gravitational/teleport/api v0.0.0-20240221014947-7cec562a5fe2
 	github.com/onsi/ginkgo/v2 v2.17.1
 	github.com/onsi/gomega v1.32.0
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.21
 
 require (
 	github.com/giantswarm/apiextensions-application v0.6.1
-	github.com/giantswarm/clustertest v0.17.0
+	github.com/giantswarm/clustertest v0.17.1-0.20240418092820-fb25dad2e6c8
```
nprokopic, line 7: Will update when https://github.com/giantswarm/clustertest/pull/166 is merged and released.

internal/upgrade/upgrade.go
```diff
@@ -93,6 +98,49 @@ func Run(cfg *TestConfig) {
...
+			}
+
+			if !controlPlaneRollingUpdateStarted {
+				Skip("Control plane nodes rolling update is not happening")
+			}
+
+			mcClient := state.GetFramework().MC()
+			Eventually(
+				wait.IsKubeadmControlPlaneConditionSet(state.GetContext(), mcClient, cluster.Name, cluster.GetNamespace(), kubeadm.MachinesSpecUpToDateCondition, corev1.ConditionTrue, ""),
+				30*time.Minute,
```
AverageMarcus, line 139: This is a loooonnnnggggg time to wait in the instance that something went wrong. Is this really a realistic timeframe for this?

nprokopic: Probably not really needed this much. In my tests it was around 10-15 minutes. Will put 20 minutes here (15 is a bit of a stretch).

AverageMarcus: Sounds good. We can look at decreasing it in the future if we manage to improve the product to be quicker.

internal/upgrade/upgrade.go
```diff
@@ -93,6 +98,49 @@ func Run(cfg *TestConfig) {
...
 		})
 
+		It("successfully finishes control plane nodes rolling update if it is needed", func() {
+			// Check MachinesSpecUpToDate condition on KubeadmControlPlane. Repeat the check 5 times, with some waiting time,
+			// so Cluster API controllers have time to react to upgrade (it is usually instantaneous).
+			numberOfChecks := 5
+			waitBetweenChecks := 5 * time.Second
+			controlPlaneRollingUpdateStarted := false
+
+			for i := 0; i < numberOfChecks; i++ {
```
mcharriere, line 108: why not using `Eventually` instead?

internal/upgrade/upgrade.go
nprokopic: Yeah, just not very optimistic here, at least not ATM, as most of that time is on the provider, e.g. getting new machine on AWS, and then on CAPI, e.g. CAPI kubeadm control plane provisioning the nodes.

internal/upgrade/upgrade.go
nprokopic: Can I use `Eventually` and then if it fails just `Skip`?

Because here the rolling update might start, but it doesn't have to (e.g. in case of apps update, or some other simpler update that does not roll the control plane nodes).

So the check here checks multiple times if the rolling update has started (when it starts most probably the first check will catch it, as CAPI reacts instantly and you see the needed condition values there), and if it hasn't, then it just skips the test (assuming that rolling update is not happening).

mcharriere: ah! okok. I understand. I'm not sure if there is a way to do that with Eventually.

internal/upgrade/upgrade.go
nprokopic: Updated timeout to 20 minutes.

AverageMarcus (approved):

PR #287, opened 2024-05-08
Add test for CAPA cluster in Cilium ENI mode

AndiDog:
### What this PR does

Towards https://github.com/giantswarm/roadmap/issues/2563

To ensure we're testing the right thing, I've added a basic "is this really ENI mode" check (pod IPs).

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Running a single test suite with teleport test enabled (e.g. the `capa` `standar
 ```sh
 kubectl get secrets teleport-identity-output -n tekton-pipelines --template='{{.data.identity}}' | base64 -D > teleport-identity-file.pem
 
-E2E_KUBECONFIG=/path/to/kubeconfig.yaml TELEPORT_IDENITTY_FILE=/path/to/teleport-identity-file.pem -v -r ./providers/capa/standard
+E2E_KUBECONFIG=/path/to/kubeconfig.yaml TELEPORT_IDENTITY_FILE=/path/to/teleport-identity-file.pem -v -r ./providers/capa/standard
 ```
 
 Running with Docker:
--- a/providers/capa/cilium-eni-mode/capa_test.go
+++ b/providers/capa/cilium-eni-mode/capa_test.go
@@ -13,7 +13,7 @@ import (
 	"github.com/giantswarm/cluster-test-suites/internal/state"
 )
 
-var _ = Describe("Cilium ENI mode tests", Ordered, func() {
+var _ = Describe("Cilium ENI mode tests", func() {
 	common.Run(&common.TestConfig{
 		AutoScalingSupported:         true,
 		BastionSupported:             false,
--- a/providers/capa/cilium-eni-mode/capa_test.go
+++ b/providers/capa/cilium-eni-mode/capa_test.go
@@ -1,12 +1,19 @@
 package cilium_eni_mode
 
 import (
+	"context"
+	"fmt"
+	"strings"
+
 	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
 
 	"github.com/giantswarm/cluster-test-suites/internal/common"
+	"github.com/giantswarm/cluster-test-suites/internal/state"
 )
 
-var _ = Describe("Common tests", func() {
+var _ = Describe("Cilium ENI mode tests", Ordered, func() {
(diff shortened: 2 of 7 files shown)


whites11 (approved):

AndiDog:
Trying again – at least the CAPA tests should succeed but several tests and cluster deletion timeout are flaky

/run cluster-test-suites

README.md
```diff
@@ -43,7 +43,7 @@ Running a single test suite with teleport test enabled (e.g. the `capa` `standar
 ```sh
 kubectl get secrets teleport-identity-output -n tekton-pipelines --template='{{.data.identity}}' | base64 -D > teleport-identity-file.pem
 
-E2E_KUBECONFIG=/path/to/kubeconfig.yaml TELEPORT_IDENITTY_FILE=/path/to/teleport-identity-file.pem -v -r ./providers/capa/standard
```
AverageMarcus, line 46: Good catch 😁

AndiDog:
Trying again after rebase, just to be sure

 /run cluster-test-suites

AndiDog:
The Crossplane providers are missing on `grizzly` – there's no CRD `SecurityGroup*` which we use to create necessary cloud resources for Cilium ENI mode.

providers/capa/cilium-eni-mode/capa_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/giantswarm/cluster-test-suites/internal/common"
+	"github.com/giantswarm/cluster-test-suites/internal/state"
+)
+
+var _ = Describe("Cilium ENI mode tests", Ordered, func() {
```
AverageMarcus, line 16: Why is this test suite set as an ordered suite?

AndiDog: I thought I need it to run

```
It("assigns IP addresses from secondary VPC CIDR to pods", func() {
```

last, i.e. right before the suite destroys the WC. If I don't do this, wouldn't it be possible that these assertions run before `common.Run` i.e. when the WC isn't ready yet? I could be wrong though 😉!

AndiDog:
Crossplane AWS providers were installed/upgraded via https://github.com/giantswarm/giantswarm-management-clusters/pull/617 and https://github.com/giantswarm/giantswarm-management-clusters/pull/618.

/run cluster-test-suites

providers/capa/cilium-eni-mode/capa_test.go
AverageMarcus: Setting it to ordered does mean that if any of the test cases fail (e.g. a flake) it doesn't attempt to run the rest of the test cases that come after it. Is that something you want? Normally (with the standard test suites) we want to attempt to run all test cases and then report back on all the cases that failed to save needing to re-run to discover more.

AndiDog: Hmm, I understood from [ginkgo documentation](https://onsi.github.io/ginkgo/#ordered-containers) that `Ordered` ensures that `It` blocks run in order, and thought we need that here since the common stuff (incl. cluster creation and health check) must definitely come first before I can check some other details about the running workload cluster. Are you sure about the behavior you described above?

AverageMarcus: We don't run tests in parallel so the order is already confirmed. I'm sure about the behaviour and you can see it happen in your previous test run here: https://tekton.ci.giantswarm.io/#/namespaces/tekton-pipelines/pipelineruns/pr-cluster-test-suites-287-cluster-test-suitesz8csx?pipelineTask=run-tests&step=run-tests&taskRunName=pr-cluster-test-suites-287-cluster-test-suitesz8csx-run-tests-0

AndiDog: Okay, I'm trying one test run without the `Ordered` flag

AverageMarcus: Just to be clear, I'm ok with it having `Ordered` if you think it adds benefit. Just as long as you're aware of the impact

AndiDog:
Looks fine. I'll merge `main` and force-merge this, since no CAPA tests are failing.

PR #305, opened 2024-05-15
increase kubeadmcontrolplane rollout timeout from 20min to 30min

vxav:
### What this PR does

CAPVCD is slower than other providers and always times out when upgrading the CP nodes.
I tested manually and babysat a pipelinerun, it all works fine, it's just slow.

### Checklist

### Trigger e2e tests

/run cluster-test-suites


ericgraf (approved):
Increasing timeout is not a great solution we always rather find the root cause of the slow down and fix that issues before it is forgottent about. 
That being said we won't block this PR .

PR #330, opened 2024-06-11
add-capa-china

calvix:
towards https://github.com/giantswarm/roadmap/issues/3482

add test sutie for CAPA china, based and china provider

In china, the DNS seems to be slower to propagate and caused issues with SSL cert for the helloworld app. The cert-manager failed to resolve DNS for helloworld application and the cert immediately went into an invalid state which cert-manager isn't retrying (it is but after 1 hour and that is way too long for e2e). To overcome this issue I introduced a DNS test that wait until the wildcard ingress DNS is resolvable and only after that deploys helloworld app.

/run cluster-test-suites


calvix:
```
Ran 27 of 29 Specs in 3278.796 seconds
SUCCESS! -- 27 Passed | 0 Failed | 0 Pending | 2 Skipped
PASS

Ginkgo ran 1 suite in 54m50.619717634s
Test Suite Passed
```

It seems to be working now, only concern is  that I saw a out of 5 tris that deployed apps needed like 5 more minutes to pass

calvix:
another successful run
```
[AfterSuite] PASSED [1252.859 seconds]

Ran 27 of 29 Specs in 3821.291 seconds
SUCCESS! -- 27 Passed | 0 Failed | 0 Pending | 2 Skipped
PASS

Ginkgo ran 1 suite in 1h3m56.656450459s
Test Suite Passed
```

assets/storage/storage.go
```diff
@@ -33,7 +33,7 @@ spec:
     runAsUser: 1001
   containers:
     - name: pvc-test-container
-      image: nginxinc/nginx-unprivileged
+      image: gsoci.azurecr.io/giantswarm/nginx-unprivileged
```
calvix, line 36: `docker.io` doesn't work in china at all

internal/common/metrics.go
```diff
@@ -178,7 +178,7 @@ func runTestPod(mcClient *client.Client, podName string, ns string) error {
 			Containers: []corev1.Container{
 				{
 					Name:  "test",
-					Image: "alpine:latest",
+					Image: "gsoci.azurecr.io/giantswarm/alpine:latest",
```
calvix, line 181: `docker.io` doesn't work in china at all

AverageMarcus (approved):

AverageMarcus:
Has the kubeconfig been updated in `tekton-resources`? If so, we should be able to run these tests in Tekton.

calvix:
> Has the kubeconfig been updated in `tekton-resources`? If so, we should be able to run these tests in Tekton.

https://github.com/giantswarm/tekton-resources/pull/358

Yes, just waiting for my last test to finish and will try to run a full test with `/run cluster-test-suites`

AverageMarcus:
You can use `/run cluster-test-suites TEST_SUITES=./providers/capa/china` to only run the new tests if you want.

AverageMarcus:
Whoops! 

I'm an idiot! I was testing something for releases and accidentally disabled the tests 🤦

AverageMarcus:
/run cluster-test-suites TARGET_SUITES=./providers/capa/china

calvix:
> /run cluster-test-suites TARGET_SUITES=./providers/capa/china
@AverageMarcus 
OK, that looks good, should we also run the test for all providers to see if my changes didn't break something else?

AverageMarcus:
Yeah lets do that to be sure but I think we're good :)

I have seen some flakey providers this morning though so as long as the majority of providers pass I think we'll be good to merge this in.

/run cluster-test-suites

calvix:
`[It] all default apps are deployed without issues` failed but all the apps were eventually deployed so everything else passed. I think China will need a longer timeout for this test like 15 more minutes.

Should I somehow make that configurable or should it be increased for all tests?

AverageMarcus:
15 minutes just for default apps to be installed is already longer than I'd like 😞 

Do you have any idea on what might be slower for China when installing these?

calvix:
> 15 minutes just for default apps to be installed is already longer than I'd like 😞
> 
> Do you have any idea on what might be slower for China when installing these?

Yeah its `app operator/chart-operator` pulling stuff from GitHub, it's quite slow

AverageMarcus:
Oh that's very annoying. 😞 

I haven't looked but can we maybe refactor it a bit to use the timeout from the `state.GetContext()`? Then we can set it differently for China (maybe).

calvix:
> Oh that's very annoying. 😞
> 
> I haven't looked but can we maybe refactor it a bit to use the timeout from the `state.GetContext()`? Then we can set it differently for China (maybe).

Yeah I have no idea how to do that, I tried a few things  but the `state.GetContext()` is returning nil when I tried to call it in `capa_test.go` or `capa_suite_test.go`

AverageMarcus:
Ok. I don't want to hold this PR up too much. 

For now just raise the timeout but would you mind adding an issue to our board to take a look at making this configurable in the future?

calvix:
> Ok. I don't want to hold this PR up too much.
> 
> For now just raise the timeout but would you mind adding an issue to our board to take a look at making this configurable in the future?

ACK, thanks

and sorry for giving you more work :see_no_evil:

AverageMarcus:
It's fine :)

calvix:
Everything except `CAPVCD` passed and that failed on cluster creation so not related to this change. 

Merging

PR #335, opened 2024-06-13
Fixes unnecessary acronyms

JosephSalisbury:
/run cluster-test-suites

### DIFF
--- a/internal/common/basic.go
+++ b/internal/common/basic.go
@@ -35,11 +35,11 @@ func runBasic() {
 			}
 		})
 
-		It("should be able to connect to MC cluster", FlakeAttempts(3), func() {
+		It("should be able to connect to management cluster", FlakeAttempts(3), func() {
 			Expect(state.GetFramework().MC().CheckConnection()).To(Succeed())
 		})
 
-		It("should be able to connect to WC cluster", FlakeAttempts(3), func() {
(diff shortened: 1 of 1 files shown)


AverageMarcus (approved):

internal/common/basic.go
```diff
@@ -33,11 +33,11 @@ func runBasic() {
 			}
 		})
 
-		It("should be able to connect to MC cluster", FlakeAttempts(3), func() {
+		It("should be able to connect to management cluster", FlakeAttempts(3), func() {
```
QuentinBisson, line 36:
```suggestion
		It("should be able to connect to the management cluster", FlakeAttempts(3), func() {
```

internal/common/basic.go
```diff
@@ -33,11 +33,11 @@ func runBasic() {
...
 			Expect(state.GetFramework().MC().CheckConnection()).To(Succeed())
 		})
 
-		It("should be able to connect to WC cluster", FlakeAttempts(3), func() {
+		It("should be able to connect to workload cluster", FlakeAttempts(3), func() {
```
QuentinBisson, line 40:
```suggestion
		It("should be able to connect to the workload cluster", FlakeAttempts(3), func() {
```

PR #339, opened 2024-06-14
Improve CAPA upgrade tests reliability and reorder common tests

nprokopic:
#### Changed

- Split common `basic` tests into `cluster basic` and `workload basic`.
- Reorder common tests, so instead of running `default apps` and then `basic`, now it first runs `cluster basic`, then `default apps`, and then `workload basic`.
- Run all common tests before the upgrade in the CAPA upgrade suite.
- Increase node pool minSize from 2 to 3 in the CAPA upgrade suite, as we have noticed that 2 nodes are not enough for all the Pods to be scheduled after the upgrade, and cluster-autoscaler does not scale up the cluster quickly enough for the test to pass.

#### Removed

- Remove node checks in the CAPA upgrade suite, because the nodes are checked in the common tests before and after the upgrade.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Split common `basic` tests into `cluster basic` and `workload basic`.
+- Reorder common tests, so instead of running `default apps` and then `basic`, now it first runs `cluster basic`, then `default apps`, and then `workload basic`.
+- Run all common tests before the upgrade in the upgrade suite.
+- Increase node pool minSize from 2 to 3 in the CAPA upgrade suite, as we have noticed that 2 nodes are not enough for all the Pods after teh upgrade, and cluster-autoscaler does not scale up the cluster quickly enough for the test to pass.
+
+### Removed
+
+- Remove node checks in the upgrade test, because the nodes are checked in the common tests before and after the upgrade.
+
 ## [1.46.0] - 2024-06-13
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
-- Remove node checks in the upgrade test, because the nodes are checked in the common tests before and after the upgrade.
+- Remove node checks in the CAPA upgrade suite, because the nodes are checked in the common tests before and after the upgrade.
 
 ## [1.46.0] - 2024-06-13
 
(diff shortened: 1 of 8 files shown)


Gacko (approved):

nprokopic:
Looks like I missed something, because in CAPA tests, after splitting `basic` tests into `cluster basic` and `workload basic`, the basic tests are not running at all 😕 will check why

nprokopic:
> Looks like I missed something, because in CAPA tests, after splitting `basic` tests into `cluster basic` and `workload basic`, the basic tests are not running at all 😕 will check why

Scratch that, it runs them, but it runs them after App tests.

AverageMarcus (changes requested):
I'd like to hold off on merging this until I have more time to review it next week as it's quite a large change being made in the middle of several changes already being made to support Releases and the in-progress transition to consolidated default-apps. Things are getting complicated / messy _very_ quickly. 😞

nprokopic:
> > Looks like I missed something, because in CAPA tests, after splitting `basic` tests into `cluster basic` and `workload basic`, the basic tests are not running at all 😕 will check why
> 
> Scratch that, it runs them, but it runs them after App tests.



> > Looks like I missed something, because in CAPA tests, after splitting `basic` tests into `cluster basic` and `workload basic`, the basic tests are not running at all 😕 will check why
> 
> Scratch that, it runs them, but it runs them after App tests.

Oooh are the tests ran alphabetically? 🤔

AverageMarcus:
Shouldn't be I don't think. The Upgrade tests are explicitly marked as `Ordered`.

nprokopic:
Oh wait, upgrade test is ordered, but the Standard test is not https://github.com/giantswarm/cluster-test-suites/blob/30351f2431fd01bc1f055a3bb5d988860ff68277/providers/capa/standard/capa_test.go#L9

AverageMarcus:
Correct. Thats expected. 

Eventually we want to run tests in parallel to speed things up but it turns out thats _really_ hard when things rely on the cluster being "ready" 😆 

The upgrade tests have a very explicit order that needs to be performed - create, upgrade, test - so must not be parallel in the future.

nprokopic:
> Correct. Thats expected.
> 
> Eventually we want to run tests in parallel to speed things up but it turns out thats _really_ hard when things rely on the cluster being "ready" 😆
> 
> The upgrade tests have a very explicit order that needs to be performed - create, upgrade, test - so must not be parallel in the future.

Looks like there are more variables here then it seemed initially. Now when I think of it, I would close this PR and open a new one that just:
- increases min nodes from 2 to 3 for CAPA upgrade test, and
- remove unnecessary nodes check in upgrade test after the upgrade is done, because we do that check already as we run all common tests after the upgrade

P.S. The second bullet above it not needed for any fix, just more because I remembered this redundant nodes check now when debugging these issues here.

nprokopic:
Closing this in favour of smaller PR here https://github.com/giantswarm/cluster-test-suites/pull/342 that is just focused on improving the reliability of CAPA upgrade suite (without affecting other providers).

PR #357, opened 2024-06-24
Test for hello-world ingress Certificate ready

AverageMarcus:
### What this PR does

Towards: https://github.com/giantswarm/giantswarm/issues/31013

Adds a new test case to the hello-world tests to check that the associated Certificate is created and is ready.

Note: As we're now referencing types from cert-manager in multiple places it doesn't make sense to keep using the `unstructured` type. I won't do it as part of this PR but I plan to update `clustertest` to include the cert-manager schema in the kube client it creates that can then be used in these tests.

Example output:

```
Common tests hello world should have a ready Certificate generated
/Users/marcus/Code/GiantSwarm/cluster-test-suites/internal/common/hello.go:206
  {"level":"info","ts":"2024-06-24T09:32:27+01:00","msg":"Checking for certificate 'hello-world-tls' in namespace 'giantswarm'"}
  {"level":"info","ts":"2024-06-24T09:32:27+01:00","msg":"Found status.condition with type 'Ready' and status 'True' in Certificate 'hello-world-tls'"}
• [0.051 seconds]
```

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added a test case to ensure the hello-world ingress has a ready Certificate
+
 ## [1.49.0] - 2024-06-21
 
 ### Changed
--- a/internal/common/hello.go
+++ b/internal/common/hello.go
@@ -248,7 +248,7 @@ func runHelloWorld(externalDnsSupported bool) {
 
 				logger.Log("Certificate '%s' is not Ready - '%s'", certificateName, conditionMessage)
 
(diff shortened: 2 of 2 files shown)


AverageMarcus:
```
Common tests hello world should have a ready Certificate generated
/app/internal/common/hello.go:206
  {"level":"info","ts":"2024-06-24T09:34:32Z","msg":"Checking for certificate 'hello-world-tls' in namespace 'giantswarm'"}
  {"level":"info","ts":"2024-06-24T09:34:32Z","msg":"Certificate 'hello-world-tls' is not Ready - 'Issuing certificate as Secret does not exist'"}
• [0.041 seconds]
------------------------
```

🤦

PR #388, opened 2024-07-08
Added failure handling for not having expected number of ready replicas

AverageMarcus:
### What this PR does

Towards: https://github.com/giantswarm/giantswarm/issues/31013

Makes use of new failure handlers to add extra debug information when Deployments, DaemonSets or StatefulSets fail to become ready before the test timeout.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


mcharriere (approved):

AverageMarcus:
😆  The CAPA China test decided to fail in exactly the right way to actually show this change in action. Amazing! 

Adding the skip label as this shows the change works as intended.

PR #398, opened 2024-07-15
Re-enable `capa-private` tests because EC2 user data limit issue should be solved now that we store Ignition bootstrap data in S3

AndiDog:
### What this PR does

https://github.com/giantswarm/roadmap/issues/3442 should've fixed this. See https://github.com/giantswarm/roadmap/issues/3442.

### Checklist

- [x] Update changelog in CHANGELOG.md.

### Trigger e2e tests

(done via comment, see below)

PR #412, opened 2024-07-24
Release v1.62.0

taylorbot:


AndiDog (approved):
Contains only #398 where the private tests worked fine, not re-running the pipeline in this PR

PR #507, opened 2024-10-11
Add debugging of Cluster CR on standup failure

AverageMarcus:
### What this PR does

Include some extra debugging of the WCs Cluster CR status when the cluster fails to standup.

Example output:

```
  {"level":"info","ts":"2024-10-11T16:35:34+01:00","msg":"Attempting to get debug info for Cluster App"}
  {"level":"info","ts":"2024-10-11T16:35:34+01:00","msg":"Cluster App status: AppVersion='', Version='0.61.0', ReleaseStatus='deployed', ReleaseReason='', LastDeployed='2024-10-11 16:16:02 +0100 BST'"}
  {"level":"info","ts":"2024-10-11T16:35:34+01:00","msg":"Getting events for the Cluster App"}
  {"level":"info","ts":"2024-10-11T16:35:34+01:00","msg":"No events found for Cluster App"}
  {"level":"info","ts":"2024-10-11T16:35:34+01:00","msg":"Getting Cluster CR for the Cluster App"}
  {"level":"info","ts":"2024-10-11T16:35:34+01:00","msg":"Cluster status: Phase='Provisioned', InfrastructureReady='true', ControlPlaneReady='false', FailureReason='', FailureMessage=''"}
  {"level":"info","ts":"2024-10-11T16:35:34+01:00","msg":"Cluster condition with type 'Ready' and status 'False' - Message='Scaling up control plane to 3 replicas (actual 1)', Reason='ScalingUp', Last Occurred='2024-10-11 16:16:44 +0100 BST'"}
  {"level":"info","ts":"2024-10-11T16:35:34+01:00","msg":"Cluster condition with type 'ControlPlaneInitialized' and status 'False' - Message='Waiting for control plane provider to indicate the control plane has been initialized', Reason='WaitingForControlPlaneProviderInitialized', Last Occurred='2024-10-11 16:16:03 +0100 BST'"}
  {"level":"info","ts":"2024-10-11T16:35:34+01:00","msg":"Cluster condition with type 'ControlPlaneReady' and status 'False' - Message='Scaling up control plane to 3 replicas (actual 1)', Reason='ScalingUp', Last Occurred='2024-10-11 16:16:44 +0100 BST'"}
  {"level":"info","ts":"2024-10-11T16:35:34+01:00","msg":"Cluster condition with type 'InfrastructureReady' and status 'True' - Message='', Reason='', Last Occurred='2024-10-11 16:16:18 +0100 BST'"}
```

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/internal/suite/setup.go
+++ b/internal/suite/setup.go
@@ -6,6 +6,8 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 
 	cb "github.com/giantswarm/cluster-standup-teardown/pkg/clusterbuilder"
 	"github.com/giantswarm/cluster-standup-teardown/pkg/standup"
@@ -13,6 +15,8 @@ import (
 	"github.com/giantswarm/clustertest/pkg/client"
 	"github.com/giantswarm/clustertest/pkg/logger"
 	"github.com/giantswarm/clustertest/pkg/utils"
+	capi "sigs.k8s.io/cluster-api/api/v1beta1"
+	cr "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/giantswarm/cluster-test-suites/internal/state"
 )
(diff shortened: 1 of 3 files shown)


AverageMarcus:
This change can be seen in action on this failing test - https://tekton.ci.giantswarm.io/#/namespaces/tekton-pipelines/pipelineruns/pr-cluster-test-suites-507-cluster-test-suites8z5md?pipelineTask=run-tests&taskRunName=pr-cluster-test-suites-507-cluster-test-suites8z5md-run-tests-9&step=run-tests

AverageMarcus:
/run cluster-test-suites TARGET_SUITES=./providers/capv/standard,./providers/capvcd/standard

internal/suite/setup.go
```diff
@@ -70,13 +74,41 @@ func Setup(isUpgrade bool, clusterBuilder cb.ClusterBuilder, clusterReadyFns ...
...
-				}
-				if len(events.Items) == 0 {
-					logger.Log("No events found for Cluster App")
+				} else {
+					if len(events.Items) == 0 {
+						logger.Log("No events found for Cluster App")
+					}
+					for _, event := range events.Items {
+						logger.Log("Event: Reason='%s', Message='%s', Last Occurred='%v'", event.Reason, event.Message, event.LastTimestamp)
+					}
```
Gacko, line 83: Just a question, because I'm not 100% into the code: Do we need the `else` here or could we also just omit it for readability? At least I don't get the difference as long as events is never `nil`. And if it would, checking only `err` doesn't feel sufficient to me.

AverageMarcus: More for consistency with the other checks in this function. 

But because we want it to log out in both cases but not both today. If there is an error we want to let the user know about it but if there wasn't an error we want to display logs based on what we fetched.

Gacko: Ah, right! Makes sense! Otherwise we would end up with 2 lines of log for the same reason.

PR #540, opened 2024-11-08
Added SetResponsibleTeam to associate teams with test cases

AverageMarcus:
### What this PR does

Towards: https://github.com/giantswarm/giantswarm/issues/32040

This change allows specific teams to be marked as "owning" a specific set of test cases that can then be responsible to investigate failures and regressions. This change introduces the framework to annotate test cases with teams and populates those team names (and associated Slack group handles) into a report JSON that can later be used by the Tekton Pipelines (not part of this PR, will come after this is released).

Be aware that not all test cases can have a single owning team as they are generic and cover cross-team features - for example checking that all Apps are installed. For cases like this we will attempt to interpret the failure and try and work out the relevant team but that will also come in a future PR.

A list of the teams responsible for failing tests can be retrieved using something like the following:
```sh
cat out/test-results.json | \
  jq -r '[.[].SpecReports[] | select(.State == "failed") | ((.ReportEntries // [])[] | select(.Name=="TEAM") | .Value.Representation)] | unique'
```

The slack handles for the teams with failing tests can be retrieved with:
```sh
cat out/test-results.json | \
  jq -r '[.[].SpecReports[] | select(.State == "failed") | ((.ReportEntries // [])[] | select(.Name=="TEAM_ID") | .Value.Representation)] | unique'
```

### Checklist

### Trigger e2e tests

/run cluster-test-suites


njuettner (approved):
👍🏻

QuentinBisson (approved):

internal/common/scale.go
```diff
@@ -33,6 +34,8 @@ func runScale(autoScalingSupported bool) {
 				Skip("autoscaling is not supported")
 			}
 
+			helper.SetResponsibleTeam(helper.TeamPhoenix)
```
fiunchinho, line 37: I believe this should be tenet, [because CAS is owned by tenet](https://github.com/giantswarm/cluster-autoscaler-app/blob/main/CODEOWNERS#L2).

AverageMarcus: Yeah I wasn't sure about this one as auto-scaling is only a CAPA feature. I'm happy to move it to our team but I'm not sure if it would actually be changes we make that might cause this to fail. What do you think?

AverageMarcus:
/run cluster-test-suites TARGET_SUITES=./providers/capa/china

internal/common/scale.go
AverageMarcus: I'm going to switch it to Tenet for now just to unblock this PR but please could you discuss it in Phoenix to see if y'all think it makes sense or not? :)

TheoBrigitte (approved):

AverageMarcus:
Test failures are flakes and unrelated to these changes. Adding skip label.

Gacko (approved):

fiunchinho (approved):

fhielpos (approved):

PR #541, opened 2024-11-13
Skip capv-on-capa and capv-on-capz tests until we can figure out VPN routing to the new vSphere provider

glitchcrab:
### What this PR does

towards https://github.com/giantswarm/giantswarm/issues/32054

### Checklist


glitchcrab:
ping @AverageMarcus is this the right way to disable the multiprovider tests for the time being? we need to kill the old provider ASAP to stop paying for two and the AWS and Azure VPNs are blocking this at the mo

vxav (approved):

PR #621, opened 2025-02-19
Metrics: Replace deprecated `pod_scheduling_duration_seconds` by `scheduling_attempt_duration_seconds`.

Gacko:
### What this PR does

It replaces the deprecated `pod_scheduling_duration_seconds` metric by the `scheduling_attempt_duration_seconds`.

`pod_scheduling_duration_seconds` is no longer available by default in Kubernetes v1.30.x and there needs a successor, so we can still test scheduler metrics.

These metrics are not equal, but that does not matter here, as we are only checking if they exist.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


Gacko:
/run cluster-test-suites TARGET_SUITES=./providers/capv/standard

Gacko:
I think something's wrong with CAPV atm. All the other tests passed and this change is not related to CAPV cluster standup, so I'm going to skip CI. 🙄

PR #664, opened 2025-04-30
Disable node-termination-handler in CAPA tests

iuriaranda:
### What this PR does

NTH may frequently restart during cluster creation due to IRSA operator being slow when creating the cluster openid provider in AWS. This makes the E2E tests fail regularly, although we consider it an expected situation. Until there's more granular control on the Pod restarting test conditions we'll disable NTH during E2E tests

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix linting issues.
 
+### Changed
+
+- Disable node-termination-handler in CAPA tests. See [giantswarm/giantswarm#32656](https://github.com/giantswarm/giantswarm/issues/32656)
+
 ## [1.87.2] - 2025-03-08
 
 ### Changed
--- a/providers/capa/china/test_data/cluster_values.yaml
+++ b/providers/capa/china/test_data/cluster_values.yaml
@@ -1,1 +1,4 @@
 # Values provided here merge on top of the default values found in https://github.com/giantswarm/cluster-standup-teardown
+global:
+  providerSpecific:
+    nodeTerminationHandlerEnabled: false # https://github.com/giantswarm/giantswarm/issues/32656
--- a/providers/capa/cilium-eni-mode/test_data/cluster_values.yaml
+++ b/providers/capa/cilium-eni-mode/test_data/cluster_values.yaml
@@ -2,6 +2,7 @@
 global:
   providerSpecific:
     awsClusterRoleIdentityName: giantswarm-grizzly-wc-e2e
+    nodeTerminationHandlerEnabled: false # https://github.com/giantswarm/giantswarm/issues/32656
   connectivity:
     # These two settings enable Cilium ENI mode
     network:
--- a/providers/capa/private/test_data/cluster_values.yaml
+++ b/providers/capa/private/test_data/cluster_values.yaml
@@ -1,1 +1,4 @@
 # Values provided here merge on top of the default values found in https://github.com/giantswarm/cluster-standup-teardown
+global:
+  providerSpecific:
+    nodeTerminationHandlerEnabled: false # https://github.com/giantswarm/giantswarm/issues/32656
--- a/providers/capa/standard/test_data/cluster_values.yaml
+++ b/providers/capa/standard/test_data/cluster_values.yaml
@@ -2,3 +2,4 @@
 global:
   providerSpecific:
     awsClusterRoleIdentityName: giantswarm-grizzly-wc-e2e
+    nodeTerminationHandlerEnabled: false # https://github.com/giantswarm/giantswarm/issues/32656
--- a/providers/capa/upgrade/test_data/cluster_values.yaml
+++ b/providers/capa/upgrade/test_data/cluster_values.yaml
@@ -1,6 +1,7 @@
 global:
   providerSpecific:
     awsClusterRoleIdentityName: giantswarm-grizzly-wc-e2e
+    nodeTerminationHandlerEnabled: false # https://github.com/giantswarm/giantswarm/issues/32656
   nodePools:
     # Here we override the default values from cluster-standup-teardown and make the following changes:
     #


fiunchinho (approved):

robinkb:
For what it's worth, we will/should remove NTH when we roll out Karpenter, because Karpenter handles the same events as NTH. I don't know when that's scheduled to happen, though. But maybe we can accept some risk on NTH regressions until then if we can stabilize the E2E tests.

Gacko:
Sure, but even if we replace NTH by Karpenter, the root cause still exists - right? It would be nice to have a solution for that when introducing Karpenter.

T-Kukawka:
@Gacko root cause for NTH issues with restarts due to IRSA are very hard to fix due to that exact dependency. In phoenix we have agreed to disable this app for tests as it is most efficient way to mitigate the issues while running the e2e tests as well as mc-bootstrap. We have spent some time on trying to improve the situation but it is not trivial.

A top of that we are working on two tracks here. One is to introduce Karpenter as a NP management system and eventually customers moving towards that 'full time', and as mentioned, introduction of Karpenter would eliminate usage of NTH.
Secondly we are working also on moving IRSA to be managed via Crossplane per app that needs it, where we will eventually get rid of the irsa-operator, that has proven to bring a lot of issues with deployment ordering and failing dependencies as in this case.

Hence we believe this is the most efficient way to move forward.

Gacko:
I totally get your point, but ultimately it leads to NTH no longer being covered by tests and therefore making releases potentially broken out of the box.

We already had quite a lot of struggle when introducing NTH in the v29 releases and I really don't wanna leave it uncovered.

I'd prefer waiting for Marcus to get some input on other possible solutions.

T-Kukawka:
Happy to get an input from @AverageMarcus on this however i don't think it brings anyone any good to block this PR. We can further evaluate next PRs based on the input tho.

We will still get rid of NTH in the future, irsa-operator will still cause the app to restart 'too many times' for the e2e tests results so e2e tests will still fail and we do not have any other alternative on Phoenix side due to how the app platform works and how the components are deployed or in which order (and this won't get fixed either)

Gacko:
IIRC Rocket was facing similar issues in the past and they added a sidecar to the actual workload to check prerequisites before the actual container was about to start. This made the actual workload work on the first start.

Anyway, let's wait for Marcus' impact. I'm just not a big fan of reducing test coverage of an already flaky app.

fiunchinho:
> Anyway, let's wait for Marcus' impact. I'm just not a big fan of reducing test coverage of an already flaky app.

The `node-termination-handler-app` is not flaky though. The app works as expected, it's just how the environment works that makes the tests to fail. But the app is working fine AFAIK.
IMHO it doesn't make sense investing time on this, while we could be using that time working on the replacement that we want to have going forward. Disabling the app in the mean time make the tests less flaky, so I think we're better off disabling it.

Gacko:
Ah, I found what Rocket was confronted with and how they solved it. Actually not that far from our current situation.

Here's the issue: https://github.com/giantswarm/giantswarm/issues/32504
Here's the fix: https://github.com/giantswarm/cloud-provider-vsphere-app/pull/205

tl;dr: Containers of the vSphere CSI Node Driver were restarting too often because they were around before their CRDs. @glitchcrab added an init container to check them before the actual container started.

Gacko:
> > Anyway, let's wait for Marcus' impact. I'm just not a big fan of reducing test coverage of an already flaky app.
> 
> The `node-termination-handler-app` is not flaky though. The app works as expected, it's just how the environment works that makes the tests to fail. But the app is working fine AFAIK. IMHO it doesn't make sense investing time on this, while we could be using that time working on the replacement that we want to have going forward. Disabling the app in the mean time make the tests less flaky, so I think we're better off disabling it.

Indeed, the app is working fine as soon as it comes up. And I am happy that we're in this state! But the moment we disable the app in the tests, we can no longer verify it is still like this in future releases. It might break even if unchanged, maybe due to changes in the environment, and we wouldn't catch it in releases but only when finally deployed in customer clusters.

And then people will have a hard time to find out why it hasn't been tested and why it was disabled. And we will need to fix it after it got released and create a hot fix release for it.

I really don't wanna go down that road. 🙂

T-Kukawka:
Looking at our load right now on Phoenix side, we cannot easily switch priorities to work on the sidecar right now taking into consideration all tasks and migrations at hand. 

I am happy to put this in the backlog to improve the situation in long run, however I believe we need an immediate impact/action for the flaky tests.

Gacko:
I expected that and if it is like this, I'm fine. Just wanted to make clear that there could be a real solution other than just disabling it and that there is a real risk when disabling it.

I was quite a bit annoyed when this app got introduced, because it was introduced in `cluster-aws` while still buggy and IIRC Andreas just fixed it afterwards. And overall it's still not perfect, otherwise we wouldn't discuss here. Last but not least it depends on the environment and even if the app itself isn't getting changed, it can still break due to changes in the environment.

This is why it's really important to me to have the app covered by tests, even if flaky. So as long as Marcus isn't coming up with a nice solution implemented in the E2E Tests, I'm really tempted to rather have it flaky and know about the root cause then completely disabled and impact customer clusters after publishing a release.

AverageMarcus:
👋  I'm here, hello!

My thoughts on this...

* Tenet aren't the gatekeepers of tests, we don't decide what should and shouldn't be tested but rather empower other teams to be able to do that in the best possible way. In this instance, we have failed at that as there's no mechanism in the framework to account for this annoying-but-valid behaviour. 
* Phoenix owns their CAPA tests and are responsible for their product. If they choose to take the risk of having reduced test coverage in favour of more stable tests overall then that's a perfectly fine decision.
* I will approve this PR but I want to make sure that there is an open issue to cover removing this change in the future when this is no longer needed - either because we've implemented pre-requisite checking into the app or because we've replaced the app with Karpenter. @T-Kukawka could you please link to such an issue? 🙏 

---

Aside from this specific PR, I would like to better understand the actual issue with the App if anyone is willing to walk me through it. If I'm following correctly it's roughly that NTH relies on IRSA but it's not enough to just rely on the irsa-operator App as a dependency because that app takes a little while to actually start up and created the openid stuff, is that correct?
If so, I see two possible improvements to our product here:
1. Update irsa-operator to not be marked as ready until everything is configured. I _think_ (but not sure) this would delay the dependant apps from installing.
2. Introduce an init-container into the NTH app that waits on the environment being available in a way that's required.

AverageMarcus (approved):

Gacko:
> Aside from this specific PR, I would like to better understand the actual issue with the App if anyone is willing to walk me through it. If I'm following correctly it's roughly that NTH relies on IRSA but it's not enough to just rely on the irsa-operator App as a dependency because that app takes a little while to actually start up and created the openid stuff, is that correct? If so, I see two possible improvements to our product here:
> 
> 1. Update irsa-operator to not be marked as ready until everything is configured. I _think_ (but not sure) this would delay the dependant apps from installing.
> 2. Introduce an init-container into the NTH app that waits on the environment being available in a way that's required.

IRSA Operator is running in the MC and not a part of the Release CR / workload cluster. It's being installed via app collections.

But the moment a new cluster is being created, the IRSA Operator ensures there's _stuff_ setup in the AWS account, that is required for the AWS NTH app inside the cluster to work correctly.

Most the time IRSA Operator does create these requirements in time. As far as I remember there's a DNS entry (`irsa.<cluster-domain>`) and HTTPS endpoint involved. But sometimes and even due to other clusters stuck in deletion, the IRSA Operator takes more time to setup everything.

IIRC I even saw all the stuff being setup but the NXDOMAIN or some wrong DNS response (probably the wildcard DNS entry for the cluster domain) still being cached by the node-local DNS cache, also leading to AWS NTH app not being able to contact its IRSA endpoint.

We were facing similar issues on vSphere when we introduced the check for restarting containers in the E2E tests. @glitchcrab solved it as described [here](https://github.com/giantswarm/cluster-test-suites/pull/664#issuecomment-2841945914). I could imagine such an init container in the actual AWS NTH workload, which checks the availability of the DNS record and IRSA endpoint, would help us prevent the restarts of the container.

Gacko (approved):
Not gonna block this any longer since Marcus was able to share his opinion. 🙂

iuriaranda:
@Gacko are the tests stuck or something? They don't seem to be running

Gacko:
Yeah, we had some issues with the CI/CD cluster this morning. Should work now.

PR #684, opened 2025-06-11
Increase Restart Limit

paurosello:
### What this PR does

I am not sure if it's due to Crossplane being a little bit slower than IRSA Operator (I doubt it but I haven't really done much benchmarks here) but cluster-autoscaler is always taking 5 restarts in the Standard mode until it's working properly.

https://tekton.ci.giantswarm.io/#/namespaces/tekton-pipelines/pipelineruns/pr-cluster-aws-1146-cluster-test-suitesh8tcv?pipelineTask=run-tests&taskRunName=pr-cluster-aws-1146-cluster-test-suitesh8tcv-run-tests-3

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Increase restart limit to 5. Some applications like cluster-autoscaler depend on IRSA and need additional time to start working properly.
+
 ## [1.88.0] - 2025-05-30
 
 ### Fixed
--- a/internal/common/basic.go
+++ b/internal/common/basic.go
@@ -158,7 +158,7 @@ func runBasic() {
 		It("doesn't have restarting pods", func() {
 			Eventually(
 				wait.ConsistentWaitCondition(
-					wait.AreNoPodsCrashLooping(state.GetContext(), wcClient, 2),
+					wait.AreNoPodsCrashLooping(state.GetContext(), wcClient, 5),
 					10,
 					5*time.Second,
 				)).


njuettner (approved):
Let's try that.

Gacko (changes requested):
This also happened in China without your change regarding the IRSA Crossplane resources, but worked in the past. So I guess something broke in the meantime.

paurosello:
@Gacko yeah, we are replacing IRSA Operator for Crossplane Resources which is a pretty bug change.

There is one drawback of this which is that we can't retrieve the distribution domain until it's fully deployed (takes some minutes) as Crossplane does not expose this value until it's fully deployed. In the operator we don't have this limitation so it's a bit faster.

I currently can't see any way to improve this on the Crossplane side.

Gacko:
No, I was talking about runs of Cluster Test Suites before your change. There the Cluster Autoscaler also already failed in the last few days, but it was good like last week.

Gacko:
I triggered a Cluster Test Suites run. Currently it should be using v3.0.0 of `cluster-aws` and therefore not include your change. Let's see if it works or not.

Gacko:
Also @AverageMarcus is AFK today and I'd like to get some input from his side regarding this.

paurosello:
perfect, I am currently shaving some seconds on the Crossplane setup by optimizing the function that gathers information. I might be able to shave 1-2 minutes as it's currently being blocked until the cluster-sa secret is present.

Gacko:
Yeah, surely that's part of the answer. Let's wait for Marcus. I'm basically fine with increasing it, I just wanted to let you know that there were issues even before the IRSA Crossplane thingy and I wanted to make sure we are not just hiding something behind increasing thresholds.

Gonna run tests once more to:

/run cluster-test-suites

Gacko:
Lovely, here's one failing, and it even exceeds the 5 restarts of this PR:

https://tekton.ci.giantswarm.io/#/namespaces/tekton-pipelines/pipelineruns/pr-cluster-test-suites-684-cluster-test-suitest7ft4?pipelineTask=run-tests&taskRunName=pr-cluster-test-suites-684-cluster-test-suitest7ft4-run-tests-3&step=run-tests

paurosello:

paurosello:
> Hello I'm here 👋
> 
> Tests shouldn't be changed to fix flakey apps. That's my opinion.
> 
> Kubernetes has `initContainers` specifically for this use case - `cluster-autoscaler` and whatever else relies on IRSA should include an initContainer that checks and wait for appropriate permissions to run.
> 
> Otherwise we risk a case where we _actually_ break IRSA permissions for these apps and the tests don't catch it because the number of restarts is below the threshold (due to whatever interval period we have set). We added these tests because we had this exact problem in the past.
> 
> So my question to you - what is preventing us from handling this in the actual apps to avoid the restarts?

My thinking is that we already have functional tests (scaling up/down the cluster in this case) that should check if the applications are working.

We could fix this multiple ways like adding an innit container, but then we need another test to check if all pods are running in under 5 minutes? I know this is not the best solution and I think we have IRSA Crossplane tuned now so this might not happen again but I am not really confident having a strict no-restart policy is great when is's kind of encouraged in Kubernetes

For example externalDNS also depends on IRSA but it will not restart if there are no permissions and IMO that is even worse.

@Gacko In the test you showed the cluster-autoscaler was not able to assume the role at all for some reason, investigating that.

Closing this issue for now as I have seen multiple CrossplaneIRSA tests going right.

PR #704, opened 2025-07-03
Implement automatic cross-major upgrade testing for releases

njuettner:
Tries to address: https://github.com/giantswarm/roadmap/issues/3994

Example for upgrades:

When drafting a v30.2.0 release

```
- Predecessor: v30.1.5 (latest v30.x.x)
- Old behavior: Test v30.1.5 → v30.2.0
- New behavior: Detect same major (30 == 30), substitute with v29.3.1 → v30.2.0
```

```
- /run releases-test-suites
- should use automatic cross-major detection, meaning uses v29.3.1 → v30.2.0 (latest previous major)
```

```
- /run releases-test-suites PREVIOUS_RELEASE=25.1.2
- uses exactly 25.1.2 → v30.2.0 (whatever was specified), can skip automatic cross-major logic (meaning manual PREVIOUS_RELEASE parameter still works as an override mechanism
```

Changes in clustertest https://github.com/giantswarm/clustertest/pull/476/files


providers/capa/china/capa_suite_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 func TestCAPAChina(t *testing.T) {
-	suite.Setup(false, &capa.ChinaBuilder{})
+	suite.Setup(false, "capa", &capa.ChinaBuilder{})
```
njuettner, line 14: This is needed for the lookup in Github figuring out the latest previous major release for a specific provider

internal/suite/setup.go
```diff
@@ -143,6 +173,46 @@ func Setup(isUpgrade bool, clusterBuilder cb.ClusterBuilder, clusterReadyFns ...
 	})
 }
 
+func getProviderFromBuilder(clusterBuilder cb.ClusterBuilder) (string, error) {
```
njuettner, line 176: Alternative for fetching the provider name, before I had used `suite.Setup(false, "capa", &capa.ChinaBuilder{})` which was not ideal

PR #724, opened 2025-07-29
Docs: Hint to install `ginkgo` from inside the repository.

Gacko:


Gacko:
Yeah, I know, but I remember `ginkgo` leaving nasty messages about version mismatches. This is why I prefer making that clear.

PR #768, opened 2025-08-28
Add automatic node roll detection to upgrade tests

njuettner:
### What this PR does

Towards: https://github.com/giantswarm/roadmap/issues/4003

Extending the upgrade test suite to automatically detect if nodes are rolled during a release upgrade.

### Changes:

A new test has been added to the upgrade suite that compares the set of cluster nodes before and after the upgrade is applied.

The result (true or false) is added to the Ginkgo test report under the `NODES_ROLLED` key.

This check can be disabled by setting the `SKIP_NODE_ROLL_DETECTION` environment variable, which is used to ignore expected node rolls (e.g., in "Previous Major" upgrade tests), because there we always roll the nodes.

A PR in `tekton-resources` will follow.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add automatic node roll detection to upgrade tests. Test reports will now include whether a release upgrade caused nodes to be rolled.
+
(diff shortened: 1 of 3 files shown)


Gacko (approved):
Nice idea, hope it works out!

We should maybe think about what happens if the cloud provider decides to remove a node during the test. Couldn't we just check if the machine template has changed instead? Because this would also cause nodes to roll.

PR #787, opened 2025-09-18
Show HelmRelease status when testing cluster fails to get ready

fiunchinho:
### What this PR does

Last week, we had an issue with the cloud controller on AWS that prevented the cluster from becoming ready. It would've been useful to see the status of the cloud controller `HelmRelease` in the logs.

### Checklist

### Trigger e2e tests

/run cluster-test-suites


fiunchinho:
@njuettner can you think of a way to test this behaviour? We need a test run that fails during the `BeforeSuite`. If we can't think of a way, I'd just merge and ignore test failures, because this change has no effect on the testing itself.

njuettner:
If it's about trying this out we could do adding in `setup.go`:

```
// Add HelmRelease status debugging
logger.Log("Getting HelmRelease status for debugging")
// Call the existing helmReleaseIssues debugging function
helmReleaseDebugger := helmReleaseIssues()
helmReleaseDebugger.Handle()
```

Since this debug logging only happens on failure, the change should not affect normal test execution.

fiunchinho:
> helmReleaseDebugger := helmReleaseIssues()
> helmReleaseDebugger.Handle()

Iv'e extracted the code to try to reuse. But Still unsure on how to test this. We can't call `helmReleaseIssues` from `setup.go`, as it's not exported. Any other ideas?

fiunchinho:

mcharriere:
> Iv'e extracted the code to try to reuse. But Still unsure on how to test this. We can't call helmReleaseIssues from setup.go, as it's not exported. Any other ideas?

The handler `helmReleaseIssues` should actually be moved to the [failurehandler package](https://github.com/giantswarm/clustertest/blob/main/pkg/failurehandler) and exported. Then could be also reused when testing apps.

fiunchinho:
> The handler `helmReleaseIssues` should actually be moved to the [failurehandler package](https://github.com/giantswarm/clustertest/blob/main/pkg/failurehandler) and exported. Then could be also reused when testing apps.

I think that's a good idea. That's what I did, please take a look.

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/fluxcd/helm-controller/api v1.3.0
 	github.com/giantswarm/apiextensions-application v0.6.2
 	github.com/giantswarm/cluster-standup-teardown v1.36.0
-	github.com/giantswarm/clustertest v1.40.0
+	github.com/giantswarm/clustertest v1.40.1-0.20250925101831-d87e299c2863
```
fiunchinho, line 10: To be changed once this is released https://github.com/giantswarm/clustertest/pull/524

Gacko (approved):

mcharriere (approved):
nice :)

fiunchinho:
/run cluster-test-suites TARGET_SUITES=./providers/capa/cilium-eni-mode

PR #788, opened 2025-09-19
use eu-north-1

paurosello:
### Checklist

### Trigger e2e tests

/run cluster-test-suites


Gacko:
As already mentioned on Slack, please keep in mind that this region does not get new features and instance types as fast as others. This might be relevant for customers wanting to make use of these. Ultimately it might be hard to debug in case we need to support them.

Gacko:
Also please consider changing that in `cluster-standup-teardown` instead of every single test suite. These values here should differentiate the test cases only.

PR #982, opened 2026-04-08
feat:  Integrate crust-gather to automatically collect cluster snapshots when tests fail

fiunchinho:
## What this PR does

This PR adds `crust-gather` to our e2e tests. `crust-gather` is a tool to collect resources from a live cluster and push them to an OCI registry, so that they can be later inspected using standard tooling like `kubectl` or `k9s`.

SIG-arch recording talking about this https://drive.google.com/file/d/1bduRue-Nyj4iAusMeuJv2gZ-QqJ_l7ev/view?usp=sharing

In this PR, we add a new flag variable to save whether or not a test case has failed during the test suite execution. In the `AfterSuite` phase, at the end of the tests suite, if this flag is true (i.e. tests have failed) we collect the cluster's resources and push them to a OCI registry in Azure. For every tests execution with failing tests, we collect the test cluster's resources, and the resources from the management cluster namespace where the test workload cluster was created.

When pushing to the OCI Registry, there are different approaches that we could take regarding which repository / image name / tag to use. I decided on always using the same repository called `snapshots`, and every pipeline execution with failing tests pushes two new tags to that repository: one for the MC and another one for the WC. The user can find the pushed tags searching for `crust-gather` in the logs in the tekton UI.

I manually created some resources in Azure
- [OCI Registry called `crustgatherci `](https://portal.azure.com/#@accountsgiantswarm.onmicrosoft.com/resource/subscriptions/1be3b2e6-497b-45b9-915f-eb35cae23c6a/resourceGroups/giantswarm-container-registry/providers/Microsoft.ContainerRegistry/registries/crustgatherci/overview) 
- [Service Principal called `crustgatherci-push`](https://portal.azure.com/#view/Microsoft_AAD_RegisteredApps/ApplicationMenuBlade/~/Overview/appId/af4b943c-3ef1-4750-931d-2fd77bdf9026/isMSAApp~/false), assigning the role `AcrPush` to it. I also created a secret for this SP, which is used by tekton to be able to push.

### How to use it

After the pipeline has finished, it should've pushed the collected resources from both the MC and the WC to the container registry. Then, we need to start the server that will serve the resources to our `kubectl` / `k9s` using `kubectl crust-gather serve`. We need to download the contents from the OCI registry
```
$ mkdir -p ./snapshots

$ docker run --rm -v $(pwd)/snapshots:/workspace -w /workspace \
    ghcr.io/oras-project/oras:v1.2.0 pull \
    crustgatherci.azurecr.io/snapshots:<cluster-name>-wc \
    --username af4b943c-3ef1-4750-931d-2fd77bdf9026 --password '<password>'

$docker run --rm -v $(pwd)/snapshots:/workspace -w /workspace \
    ghcr.io/oras-project/oras:v1.2.0 pull \
    crustgatherci.azurecr.io/snapshots:<cluster-name>-mc \
    --username af4b943c-3ef1-4750-931d-2fd77bdf9026 --password '<password>'

$ kubectl crust-gather serve --archive ./snapshots &

$ kubectl --context mc get helmreleases -A
$ kubectl --context wc get pods -A
```

Now in your current kube config file you should have two new contexts called `wc` and `mc`.

#### Troubleshooting

Since `crust-gather` serves at 0.0.0.0:9095 every time, `kubectl`'s discovery cache from a previous session (with different contents) may be stale. `kubectl` caches API resource discovery in `~/.kube/cache/discovery/` per-host. When inspecting the collected resources from the MC and WC, you may hit this issue. The quick fix: clear the cache with `rm -rf ~/.kube/cache/discovery/0.0.0.0_9095`, kill the server and restart it.

#### Updating tekton resources

For the PoC I manually created the `Secret` called `crust-gather-acr-creds` in the `tekton-pipielines` namesapce with the Azure credentials required to push to the OCI Registry. 

I also modified the tekton `Task` called `run-wc-e2e-tests`. I added the `kustomize.toolkit.fluxcd.io/reconcile: disabled` annotation (so flux wouldn't revert my changes) and added the `CRUST_GATHER_REGISTRY_USERNAME` and `CRUST_GATHER_REGISTRY_PASSWORD` environment variables to the `run-tests` step.

#### Cleaning old resources

The tags being pushed to the OCI registry are never cleaned up, so they will pill up over time. We need to add some policy / automation to clean up tags after a few days: we don't need historic data here.

#### Collect resources for all kind of tests: app tests, releases tests, etc

Ideally, we would offer the same behaviour for all our tests.

#### Docs

Add a page in the intranet about how users can inspect the collected resources for a failing pipeline.

#### Serve resources directly from OCI Registry instead of user's file system

`crust-gather` can serve collected resources directly from an OCI registry, without the user having to download the contents first using `crust-gather serve --reference crustgatherci.azurecr.io/snapshots:t-ocenn43ac41j300mwg-wc`. But for the PoC I created a `Basic` OCI Registry, and the `Basic` type is very limited in the speed / concurrency it allows. So when I tried inspecting the resources using `k9s`, the experience was not great. The resources are being download as they are requested by `k9s`, meaning it was very confusing to see the list of pods or helmreleases in the cluster, as they were showing up _eventually_. That is why in the PoC I ended up using the approach of downloading the collected resources first.
Something to try is to use a better OCI Registry i.e. `Standard` type instead of `Basic`, and test if the experience is better.

#### Improving discoverability of collected resources

Currently we log the names of the tags used to push the collected resources. A user that wants to inspect the collected resources need to find these logs in the tekton UI.

I'm using Azure OCI registry because `crust-gather` has built-in support for it, and we already use them at GiantSwarm. But we could explore collecting the resources and pushing them to Github as artifacts so that it would be easier for developers to discover the collected resources?

#### Use 'record' to record test cluster from the beginning of the tests

`crust-gather` has another feature to _record_ the state of a cluster over time. Then, you can later on replay what happened to that cluster in real time. Would it be useful to record a cluster from the beginning of the pipeline? It would mean always record, and then only push if tests have failed. But I don't know if it would be useful, because tests take a long time to execute, and the user would have to wait in real time for certain events to happen.


fiunchinho:

fiunchinho:

fiunchinho:

fiunchinho:

fiunchinho:

fiunchinho:

fiunchinho:

fiunchinho:

fiunchinho:

fiunchinho:

fiunchinho:

fiunchinho:

fiunchinho:

fiunchinho:

fiunchinho:

fiunchinho:
In the comment that we write in the PR when tests fail, we could add the steps to explore the collected resources from the test cluster.

or4dx:

internal/suite/setup.go
```diff
@@ -49,19 +60,76 @@ func WithExtraClusterValues(fn func() (string, error)) Option {
 	return func(o *Options) { o.ExtraClusterValuesFn = fn }
 }
 
+
+const (
+	OpenAIAPIKeySecretNamespace = "giantswarm"
+	OpenAIAPIKeySecretName      = "openai-api-key"
```
fiunchinho: I think this can be removed, it got removed from `main` IIRC

Dockerfile
```diff
@@ -4,6 +4,11 @@ FROM --platform=$BUILDPLATFORM golang:1.26 AS build-tests
 
 ARG TARGETOS
 ARG TARGETARCH
+ARG CRUST_GATHER_VERSION=0.15.1
+
+RUN curl -sSfL \
+    "https://github.com/crust-gather/crust-gather/releases/download/v${CRUST_GATHER_VERSION}/kubectl-crust-gather_${CRUST_GATHER_VERSION}_linux_${TARGETARCH}.tar.gz" \
+    | tar -xz -C /tmp
```
Gacko, line 10:
```suggestion

ARG CRUST_GATHER_VERSION=0.15.1
RUN curl -sSfL \
    "https://github.com/crust-gather/crust-gather/releases/download/v${CRUST_GATHER_VERSION}/kubectl-crust-gather_${CRUST_GATHER_VERSION}_linux_${TARGETARCH}.tar.gz" \
    | tar -xz -C /tmp
```

providers/capa/standard/capa_test.go
```diff
@@ -28,4 +28,5 @@ var _ = Describe("Common tests", func() {
 
 	// ECR Credential Provider specific tests
 	ecr.Run()
+
```
Gacko: Please remove this.

CHANGELOG.md
```diff
@@ -18,7 +22,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - CAPA/Upgrade: Pass organization as extra value.
-
```
Gacko, line 27: Please do not remove this.

internal/suite/setup.go
Gacko: Yes, we removed Shoot for now. I think this comes from an incorrectly solved merge conflict.

Gacko (changes requested):

njuettner (approved):

internal/suite/setup.go
```diff
@@ -49,19 +60,74 @@ func WithExtraClusterValues(fn func() (string, error)) Option {
 	return func(o *Options) { o.ExtraClusterValuesFn = fn }
 }
 
+
+const (
+
```
Gacko:
```suggestion
```

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Integrate crust-gather to automatically collect cluster snapshots (WC and MC) when tests fail, pushing them to an OCI registry for offline debugging.
+
+
```
Gacko:
```suggestion
```

Gacko (approved):

PR #986, opened 2026-04-13
Support both App-CR and HelmRelease-based default apps in e2e assertions

fiunchinho:
## Summary

- Newer cluster chart releases deploy default workload-cluster apps as Flux `HelmRelease`s instead of `App` CRs. This PR makes the e2e suite work with both styles by auto-detecting which resources exist on the management.
- **Default apps block** (`apps.go`): lists App CRs matching the default-apps selector; if none exist, `Skip()`s with a clear message — the unconditional "all HelmReleases are deployed" block already covers those releases.
- **Bundle blocks** (`apps.go`): observability-bundle and security-bundle each have an App-CR `It` and a HelmRelease `It` sibling. Each checks whether its parent resource (App or HelmRelease) exists on the MC via a cheap `Get` call and `Skip()`s immediately if not, deferring to the sibling.
- **hello.go preconditions**: cert-manager and external-dns waits use a new `WaitAppOrHelmReleaseReady` helper that polls both App CR and HelmRelease and succeeds as soon as either is Ready. No version logic needed — whichever kind the cluster chart created will be found.


fiunchinho:

fiunchinho:

fiunchinho:

njuettner (approved):

PR #1115, opened 2026-08-17
fix(crust-gather): retry snapshot collection instead of losing archives on transient failures

or4dx:
Crust-gather retries individual resource fetches forever and only writes the archive after the whole --duration window succeeds, so one stuck pod or node was enough to discard an otherwise-complete WC/MC snapshot.

- Retry the full collection up to 3 times, then fall back to one attempt without pod logs so a persistently unreachable node still yields a resource/events archive instead of nothing.
- Bump crust-gather to v0.17.0.


njuettner (approved):

or4dx:

